PR #21, opened 2018-08-17
Delete current letter on delete keydown

wishdasher:
I tested this locally. Not sure if you have more extensive testing requirements.


stevenhao:
thanks!

PR #53, opened 2018-10-15
Publish button and resize grid in compose mode

akshatb42:


stevenhao:
Looks good! Do you want to add support for non-square grids? It may also be good to show a composition’s is-published status somewhere

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

akshatb42:
Okay, now supports rows and columns separately. For the second - do you mean keep track of each variant of a published composition somewhere? Next pr lol

PR #56, opened 2019-01-08
Battle mode v0

akshatb42:
screen shot 2019-01-08 at 5 05 11 am

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -74,7 +74,7 @@ const actions = {
 
   getNextBid: (cbk) => {
     // Copying Cid logic for now...
-    const NUM_BIDS = 1000000;
+    const NUM_BIDS = 100000000;
     const bid = Math.floor(Math.random() * NUM_BIDS);
     cbk(bid);
   },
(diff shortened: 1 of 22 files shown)


src/actions.js
```diff
@@ -116,6 +124,21 @@ const actions = {
     );
   },
 
+  // TODO: this should probably be createGame and the above should be deleted but idk what it does...
```
stevenhao: p sure `createGame` is legacy at this point. i'll delete it later

src/pages/Battle.js
```diff
@@ -0,0 +1,182 @@
...
+      team: undefined,
+      games: undefined,
+      startedAt: undefined,
+      redirecting: false,
+      name: undefined,
+      players: undefined,
+    };
+    this.handleChangeName = this.handleChangeName.bind(this);
+    this.handleUnload = this.handleUnload.bind(this);
+    this.renderTeam = this.renderTeam.bind(this); // Why do I have to do this?
```
stevenhao: because of map(this.renderTeam)

PR #64, opened 2019-05-28
Add a button to .puz uploader to create unlisted (url only) puzzles

lennartjansson:
Screen Shot 2019-05-27 at 6 17 46 PM

This change adds a second button to the .puz file Upload component shown after choosing a .puz file. The second button also creates a new puzzle but it sets `private: true` so the puzzle does not appear in the public list of puzzles. After the database transaction, it shows a message to the user with a link to the /beta/play/<pid> url to start a new game with the uploaded puzzle.

### DIFF
--- a/src/components/Upload.js
+++ b/src/components/Upload.js
@@ -11,6 +11,7 @@ export default class Upload extends Component {
     this.state = {
       puzzle: null,
       recentlyCreatedPuzzleId: null,
+      unlistedCheckboxChecked: false,
     };
   }
 
@@ -18,11 +19,13 @@ export default class Upload extends Component {
     this.setState({
       puzzle: {...puzzle},
       recentlyCreatedPuzzleId: null,
+      unlistedCheckboxChecked: false,
     });
   };
 
-  create = (isPrivate = false) => {
-    const {puzzle} = this.state;
+  create = () => {
+    const {puzzle, unlistedCheckboxChecked} = this.state;
+    const isPrivate = unlistedCheckboxChecked;
     if (isPrivate) {
       puzzle.private = true;
     }
@@ -51,6 +54,12 @@ export default class Upload extends Component {
     }
   }
 
+  handleChangeUnlistedCheckbox = (e) => {
+    this.setState({
+      unlistedCheckboxChecked: e.target.checked,
+    });
+  };
+
   renderButton() {
     const {v2} = this.props;
     const {info} = this.state.puzzle || {};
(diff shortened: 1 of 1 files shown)


stevenhao:
thanks for the pr! could we make the button a bit less prominent in the page, so it doesn't affect the experience of users who don't regularly use this feature?

One option: gate based on url -- downforacross.com/?author=1 renders the same welcome page, except the upload button says "Add unlisted puzzle". 
Something like `window.location.search.indexOf('secret') !== -1`

Another option: A small checkbox that says "Unlisted" next to the upload button that toggles unlisted-ness.

src/components/Upload.js
```diff
@@ -8,18 +8,32 @@ import React, {Component} from 'react';
...
+    };
   }
 
   success = (puzzle) => {
-    this.setState({puzzle: {...puzzle}});
+    this.setState({
+      puzzle: {...puzzle},
+      recentlyCreatedPuzzleId: null,
+    });
   };
```
stevenhao:
```suggestion
get unlistedMode() { return window.location.search.indexOf('unlisted') !== -1; }
```

src/components/Upload.js
```diff
@@ -43,13 +57,36 @@ export default class Upload extends Component {
     const {type} = info || {};
     if (type) {
       return (
-        <button className={'upload--button ' + (v2 ? 'v2' : '')} onClick={this.create}>
-          {`Add to the ${type} repository`}
-        </button>
+        <div>
+          <button className={'upload--button ' + (v2 ? 'v2' : '')} onClick={(e) => this.create(false)}>
```
stevenhao:
```suggestion
          <button className={'upload--button ' + (v2 ? 'v2' : '')} onClick={(e) => this.create(this.unlistedMode)}>
```

src/components/Upload.js
```diff
@@ -43,13 +57,36 @@ export default class Upload extends Component {
...
+            {`Add puzzle to the public ${type} repository`}
```
stevenhao:
```suggestion
            {`Add puzzle to the public ${type} repository` + (this.unlistedMode ? ' (unlisted)' : '')}
```

lennartjansson:
I implemented your checkbox suggestion. Here's how it looks in practice:

Screen Shot 2019-06-01 at 5 17 53 PM

Screen Shot 2019-06-01 at 5 17 48 PM

The checkbox defaults to unchecked.

PR #81, opened 2020-02-02
Add unread chat indicator

samclearman:


src/pages/GameV2.js
```diff
@@ -194,6 +200,12 @@ export default class GameV2 extends Component {
     this.gameModel.chat(username, id, message);
   };
 
+  updateSeen = (message) => {
```
stevenhao: can we rename this to `updateSeenChatMessage`?

samclearman: done

Issue #65, opened 2019-07-01
Visibility into members in room


hjylewis:
It would be cool to see a list of people currently in the room and the ability to hover over the colored cursors and see who it is

stevenhao:
great idea!

stevenhao:
list of users is added by #83
![image](https://user-images.githubusercontent.com/1988030/77940103-0b037880-726d-11ea-9fa0-9076c147dda6.png)

hover over cursor to see who it is: still TODO, but should be relatively straightforward now

carlyrobison:
I noticed this while crosswording today. Great addition!

PR #83, opened 2020-03-30
Display user presented under chat bar

leekkww:
Changes:
- displays user presented under chat bar, with reasonable wrapping if name is too long
- moves user information (color, displayname) to a single object in the reducer
- messages now show current name corresponding to each sender (instead of the sender name at the time of message sent)

### DIFF
--- a/src/components/Chat/index.js
+++ b/src/components/Chat/index.js
@@ -260,7 +260,6 @@ export default class Chat extends Component {
     const color = this.getMessageColor(id, isOpponent);
     const users = this.props.users;
 
-    this.props.updateSeenChatMessage && this.props.updateSeenChatMessage(message);
     return (
       <div className={'chat--message' + (big ? ' big' : '')}>
         {this.renderMessageSender(users[id].displayName, color)}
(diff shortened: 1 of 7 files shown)


src/components/Chat/index.js
```diff
@@ -238,11 +259,11 @@ export default class Chat extends Component {
     const {text, senderId: id, isOpponent} = message;
     const big = text.length <= 10 && isEmojis(text);
     const color = this.getMessageColor(id, isOpponent);
+    const users = this.props.users;
 
-    this.props.updateSeenChatMessage && this.props.updateSeenChatMessage(message);
```
leekkww: this line is unrelated to this change. react was erroring saying that we shouldn't update anything while we are rendering, and the error goes away if we remove this line. lmk if you want this line back instead.

stevenhao: ah, i didn't notice this line. it should go in componentDidMount, or ideally, in a callback that triggers when the scrolls into view in the page. deleting for now is fine, i'll find a way to fix it.

stevenhao:
closes #65

PR #94, opened 2020-05-09
Fix issue where mobile chat didn't bring up a keyboard.

audreyli:
Looks like this was from when we removed the custom mobile
keyboard, but didn't make sure the default keyboard would
get rendered instead.

This just subs the mobile EditableSpan with a regular old
input. Except I added some styling to keep it more
consistent with how the current mobile chat input looks.

image


stevenhao:
great stuff, thanks!

PR #97, opened 2020-05-17
Mobile UI improvements + Material icons

cleercode:
- Size mobile toolbars for touch targets (≥ 48px)
- Remove second clue bar on mobile (is there a reason for it?)
- Replace most icons with [Material icon set](https://material.io/resources/icons/?style=baseline) to improve visuals and simplify code

desktop-home
mobile-chat
mobile-home
mobile-puzzle

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "react-flexview": "^3.0.1",
     "react-fontawesome": "^1.6.1",
     "react-helmet": "^5.2.0",
+    "react-icons": "^3.10.0",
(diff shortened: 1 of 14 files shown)


catherynl:
Love this, thanks!!
The reason for that redundant clue on the mobile page is that if you scroll down far enough the top clue will no longer be on screen... but maybe we should just make it sticky :)

cleercode:
Ah, yeah, I thought it was sticky already, but if not, I think that'd be an improvement.

PR #103, opened 2020-06-06
SweetAlert confirmation for reveal

stian8:
Alert box for _all_ Reveals


src/components/Toolbar/index.js
```diff
@@ -189,7 +190,16 @@ export default class Toolbar extends Component {
   }
 
   reveal(scopeString) {
-    this.props.onReveal(scopeString);
+    swal({
+      title: 'Are you sure you want to reveal the answer?',
+      icon: 'warning',
+      buttons: true,
+      dangerMode: true,
+    }).then((willDelete) => {
```
stevenhao, line 198: hmm, maybe this should be named something else, like `shouldReveal` (or something simple like `yes`)

stevenhao:
Thanks for the PR! Looks great, though I had one small style question

src/components/Toolbar/index.js
```diff
@@ -189,7 +190,16 @@ export default class Toolbar extends Component {
   }
 
   reveal(scopeString) {
-    this.props.onReveal(scopeString);
+    swal({
+      title: 'Are you sure you want to reveal the answer?',
```
stevenhao, line 194: Could we make the value of title vary based on `scopeString`. I think the current wording could be confusing to users; they might think they're revealing the entire grid.

stian8: Ah good point! Added and also some text :)

PR #115, opened 2020-09-13
Vim Keybindings

thkim1011:
This is more of a proof of concept (the actual code needs more cleanup work and there are some design questions that need to be addressed here), but I've managed to implement vim keybindings to the frontend. My changes adds a new action menu which allows you to switch from normal mode to vim mode. Vim mode should allow `hjkl` for navigation and `x` for deleting in normal mode, and all the original action keys from insert mode. To switch into insert mode use `i` and to switch back to normal mode use `Escpape`. If you could look at it and tell me what you think, that'd be great :).

### DIFF
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -176,16 +176,22 @@ export default class GridControls extends Component {
       k: 'up',
       l: 'right',
       x: 'delete',
-      ' ': 'space',
-      '[': 'backward',
-      ']': 'forward',
     };
 
     const {onVimNormal, onVimInsert, vimInsert, onPressEnter, onPressPeriod} = this.props;
     if (key in actionKeys) {
       this.handleAction(actionKeys[key], shiftKey);
-    } else if (!vimInsert && key in normalModeActionKeys) {
-      this.handleAction(normalModeActionKeys[key], shiftKey);
+      return true;
+    } else if (!vimInsert) {
+      if (key in normalModeActionKeys) {
+        this.handleAction(normalModeActionKeys[key], shiftKey);
+      } else if (key === 'w') {
+        this.selectNextClue(false);
+      } else if (key === 'b') {
+        this.selectNextClue(true);
+      } else if (key === 'i') {
+        onVimInsert && onVimInsert();
+      }
     } else if (key === '.') {
       onPressPeriod && onPressPeriod();
       return true;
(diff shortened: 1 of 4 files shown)


src/components/Player/GridControls.js
```diff
@@ -154,12 +155,65 @@ export default class GridControls extends Component {
...
+      l: 'right',
+      x: 'delete',
+      ' ': 'space',
+      '[': 'backward',
+      ']': 'forward',
+    };
+
+    const {onVimNormal, onVimInsert, vimInsert, onPressEnter, onPressPeriod} = this.props;
+    if (key in actionKeys) {
+      this.handleAction(actionKeys[key], shiftKey);
```
stevenhao, line 186: return true to prevent tab default behavior

src/components/Player/GridControls.js
```diff
@@ -154,12 +155,65 @@ export default class GridControls extends Component {
...
+
+    const normalModeActionKeys = {
+      h: 'left',
+      j: 'down',
+      k: 'up',
+      l: 'right',
+      x: 'delete',
+      ' ': 'space',
+      '[': 'backward',
+      ']': 'forward',
```
stevenhao, line 181: w -> selectNextClue(true), b -> selectNextClue(false) ? :D

stevenhao: and maybe dd for delete word, though that's a bit unintuitive when it's vertical i guess

src/components/Toolbar/index.js
```diff
@@ -98,6 +98,20 @@ export default class Toolbar extends Component {
     );
   }
 
+  renderKeybindMenu() {
+    const {vimMode, vimInsert} = this.props;
+    return (
+      <ActionMenu
+        label={vimMode ? `VIM ${vimInsert ? 'Insert' : 'Normal'}` : 'Normal'}
```
stevenhao, line 105: I think VIM Normal is a bit confusing at first glance. maybe strings can be
- `Normal`
- `VIM`
- `VIM (insert)`

stevenhao (commented):
i really like this, thanks for the feature! left some inline suggestions, happy to merge this in once we address those!

src/components/Player/GridControls.js
thkim1011: i noticed this bug but was having trouble figuring out why it was happening. thanks!

src/components/Player/GridControls.js
thkim1011: i added w and b (good suggestions!) but haven't added dd yet since it seems like it would take some more work (need to add some queue state to check for the first d. I could take a stab at it in another PR along with some other ideas I had e.g. use `:32` to go to clue 32.

src/components/Toolbar/index.js
thkim1011: Fixed!

thkim1011:
Hi @stevenhao , I made the changes you mentioned.

stevenhao:
Amazing :)

src/components/Player/GridControls.js
stevenhao: ooh yea, `:32` would be sweet

Issue #116, opened 2020-09-14
Getting dev backend to talk to frontend


TMiguelT:
I'm trying to set up the dev environment.

For the backend I ran the instructions detailed here: https://github.com/downforacross/downforacross.com/tree/master/server#getting-started. For the frontend I ran these commands: https://github.com/downforacross/downforacross.com#getting-started.

However, while both appear to be working, uploading a `.puz` file appears to do nothing, and I can't see it on the puzzle list. Any pointers?

stevenhao:
I need to update these docs, thanks for the reminder!

### Running frontend-only, talking to a shared backend (api-staging.foracross.com):
- `yarn start`

### Running local backend and local frontend that talk to each other:
- `yarn devbackend`
- `yarn devfrontend`

stevenhao:
Not sure why uploading a puz does nothing when you run `yarn start`, it should append to the puzzlelist in firebase (though I am actually working on migrating the entire puzzlelist & puzzle repo to postgres -- stay tuned).

TMiguelT:
Great, thanks for the clarification. Actually when I originally used `yarn start` I had an empty puzzle list, but now when I do it I'm able to see everything I uploaded.

PR #117, opened 2020-09-14
Use a tokenizing filter

### LINKED ISSUE #92: Allow searching by disjoint substrings
For example, if I want to find NYT Crosswords from April 2020, my intuition is to type `NY Times April 2020` into the search bar. This doesn't work because downforacross is searching for that entire string in each title, and not finding it because it doesn't exist within a string like `NY Times, Monday, April 27, 2020`. But it's a very common technique to split each token in the search term, and return an result that contains `NY`, `Times`, `April`,  and `2020`. I believe this would be more powerful and intuitive.

TMiguelT:
Instead of the current (confusing) filter behaviour, we instead:

* Lower case the search string
* Split it into tokens

Then, each token must be present within the (lowercase) author name, or puzzle title or it is filtered out.

This is a draft PR until I can actually test it (see #116)

Closes #92.

### DIFF
--- a/src/components/PuzzleList/PuzzleList.js
+++ b/src/components/PuzzleList/PuzzleList.js
@@ -80,10 +80,11 @@ export default class PuzzleList extends PureComponent {
     const searchMatches = search
       .toLowerCase()
       .split(/\s/)
-      .every((token) => {
-        // Each token can be either in the author's name or in the puzzle's title
-        author.includes(token) || title.includes(token);
-      });
+      .every(
+        (token) =>
+          // Each token can be either in the author's name or in the puzzle's title
+          author.includes(token) || title.includes(token)
+      );
 
     return statusFilter[status] && sizeFilter[size] && searchMatches;
   };
--- a/src/components/PuzzleList/PuzzleList.js
+++ b/src/components/PuzzleList/PuzzleList.js
@@ -73,16 +73,18 @@ export default class PuzzleList extends PureComponent {
       started: 'In progress',
     }[this.puzzleStatuses[entry.pid]];
 
-    const matches = (str, expr) => {
-      if (expr.toLowerCase() === expr) {
-        // case insensitive
-        return str.toLowerCase().indexOf(expr) !== -1;
-      } else {
-        // case sensitive
-        return str.indexOf(expr) !== -1;
-      }
-    };
-    const searchMatches = matches(entry.info.author, search) || matches(entry.info.title, search);
+    // Normalise case, and do it outside the filter loop for performance reasons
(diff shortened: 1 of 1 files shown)


src/components/PuzzleList/PuzzleList.js
```diff
@@ -73,16 +73,18 @@ export default class PuzzleList extends PureComponent {
...
+    // Normalise case, and do it outside the filter loop for performance reasons
+    const author = entry.info.author.toLowerCase();
+    const title = entry.info.title.toLowerCase();
+    // Next, we normalise the search term, and then split it to allow for non-contiguous searches
+    const searchMatches = search
+      .toLowerCase()
+      .split(/\s/)
+      .every((token) => {
+        // Each token can be either in the author's name or in the puzzle's title
+        author.includes(token) || title.includes(token);
```
stevenhao, line 85: oops, missing `return`

stevenhao:
Love this, fully support the idea! You should be able to test this without running a local backend, since the puzzle list is stored in firebase.

You can also test on the autodeploy links posted by vercel:  downforacrosscom-git-fork-tmiguelt-token-search.downforacross1.vercel.app.

TMiguelT:
I tried that, but as in the other issue, I'm not able to actually upload puzzles to the site. And the puzzle list is empty, so I can't really test out this behaviour.

TMiguelT:
And... now they're all visible. I have no idea what was going on there.

stevenhao (approved):
Looks great!

stevenhao:
image
sweet, going to merge this. maybe a follow-up can be highlighting the parts of the puzzle that matches the search term.

TMiguelT:
I suppose there's some argument for supporting search quotes, such that "LA Times" will only return that substring, and not "la" somewhere in the text and "times" somewhere in the text. But it depends on how you want the search to work: as a Google-esque plaintext search that implements features via search syntax, or as a kind of form with lots of fields that you can tweak.

This question also matters for the other thing search is missing: searching by date. Under the plaintext system you might type `date:2020-07-22`, whereas a form-based search might have a calendar popup for this purpose.

stevenhao:
Yeah, I like the idea of coming up with a more powerful syntax -- something like [gmail's operators](https://support.google.com/mail/answer/7190?hl=en).

Another extension would be to hold the filter state (the Size / Status checkboxes) in the search bar itself, so checking & unchecking would just insert/delete search terms.

TMiguelT:
That all seems quite doable. I wonder if there's a React or vanilla JS library that provides a framework for implementing an operator search like this. I can't find one, but admittedly it wouldn't be so hard to do manually.

PR #120, opened 2020-09-30
Fix URL typo in README

edemaine:
www.downforacross.com does not resolve.


stevenhao:
Thanks!

stevenhao:
in case someone stumbles upon this PR in the future -- the `www.` url also resolves now because I just updated the DNS records.

PR #129, opened 2020-12-17
Unify font styling across app and light and dark themes

pratheeknagaraj:
Makes use of the same font family in the search bar, chat, and buttons. Arial (or the default input font on a browser) is too noticeable among the various text renderings.

Makes dark theme use the same outline shape on inputs for consistency. This makes the full bounding box seen on dark theme (in line with the light theme).


stevenhao (approved):
Thank you!

PR #130, opened 2020-12-19
Start mirroring puzzle data to postgres

stevenhao:
Create puzzle now sends a POST /api/puzzle request as well.

I bit off a bit more than I could chew at first, trying to build a fully functional drop-in replacement for the existing storage system for the puzzle repository. Instead I'll take this in pieces:
1. send newly created things to both firebase and postgres [this PR]
2. migrate first read method: puzzle list to use postgres, while keeping game creation on firebase [partially implemented in this PR, but switch is not flicked yet]
3. do a backfill firebase->postgres for old puzzles (if i decide i want to keep them...)
4. finish migrating read methods to only read from postgres -- game creation can still live in frontend [partially implemented here but it was a bad idea]
5. nuke puzzles from firebase when confirming that things are stable

TS migration stuff:
- 95% of this repo has terrible js style, i'll fix that eventually... but in small pieces


server/model/game.ts
```diff
@@ -11,10 +11,21 @@ export async function getEvents(gid: string) {
...
-  user: any;
-  timestamp: Date;
-  type: any;
+  user?: string; // always null actually
+  timestamp: number;
+  type: string; // todo string literal union type
+  params: any; // todo extend GameEvent w/ specific types of game events
 }
+
+export interface InitialGameEvent extends GameEvent {
```
stevenhao, line 20: this isn't used anywhere yet

Issue #125, opened 2020-11-23
Automatic error check mode


jackofallspades98:
There should be a setting to turn on "Automatic Error Check," which would automatically mark incorrect letters with red text (like normal errors) but without needing to go through the Check-->Puzzle menu.

stevenhao:
closed in #136

PR #136, opened 2020-12-29
autocheck feature

jsbeckwith:


src/store/game.js
```diff
@@ -246,6 +246,21 @@ export default class Game extends EventEmitter {
     });
   }
 
+  updateCellAutocheck(r, c, id, color, pencil, autocheck, value) {
```
stevenhao, line 249: Another way to do this could be to add 2 events -- `updateCell`, then `checkCell`. This would simplify the reducers a bit (no need to introduce a dedicated event type for this functionality), at the expense of having a more muddled event history. Do you think you could change that?

stevenhao:
Thanks for the PR! Love the idea, it's a great "easy mode" for people new to crosswords, or perhaps trying to solve quickly :)

Requested here https://github.com/downforacross/downforacross.com/issues/125

stevenhao (approved):

src/store/game.js
jsbeckwith: Hey! That seems cleaner to me. I'm pretty new to React so doing it with a dedicated event type made the most sense to me. I just changed the implementation and will submit a new PR! Thanks!

stevenhao: Ah sorry, I misspoke! I meant that this should be done with 2 events, `updateCell` and `check` (it's a pre-existing event that you can scope to a single cell), not a new event called `checkCell`. My comment was also purely about the internal implementation of `gameModel.updateCellAutocheck`; I actually like the idea of `updateCellAutocheck` being a public method of the game model that abstracts away the underlying logic from the react components that invoke it.

So to be clear, the changes I'm suggesting is:
- no changes to lib/reducers (so revert that file to the version in master)
- new function in store/game (`updateCellAutocheck`), invoked directly from components/Game/Game.js if `autocheckMode` is true [this is what you had originally in https://github.com/downforacross/downforacross.com/pull/136/commits/0cbc78e00bc10d57177cf0b6c7efbc071e57363f]
- the function `updateCellAutocheck` is implemented via two calls to `addEvent`.
```
  updateCellAutocheck(r, c, id, color, pencil, value) {
    this.addEvent({
      timestamp: SERVER_TIME,
      type: 'updateCell',
      params: {
        cell: {r, c},
        value,
        color,
        pencil,
        id,
      },
    });
    this.addEvent({
      timestamp: SERVER_TIME,
      type: 'check',
      params: {
        scope: [{r, c}],
      },
    });
  }
```

jsbeckwith: Oh cool! Thanks for spelling it out for me. On it!

stevenhao:
closes #125

stevenhao:
Sweet, thanks for the contribution!

PR #147, opened 2021-01-21
Allow users to change their colors

### LINKED ISSUE #88: Let users change their color
Starting point: see https://github.com/downforacross/downforacross.com/pull/83

ewmson:
This PR adds in the ability for a user to change their colors.
To keep the UI clean I went with the circles picker of react-color which does limit the options that can be picked from, but is much less intimidating.

A user can change their color by simply clicking on the dot next to where they edit their name and selecting the new color.

Here is a video of that being done:

https://user-images.githubusercontent.com/6018559/105261348-5a8c9b00-5b44-11eb-8d12-d9c7decbcea8.mov

The other users are able to see the changed colors of that user:

https://user-images.githubusercontent.com/6018559/105261538-a9d2cb80-5b44-11eb-99b7-209711964fea.mov

closes #88

### DIFF
--- a/src/components/Chat/ColorPicker.tsx
+++ b/src/components/Chat/ColorPicker.tsx
@@ -1,9 +1,11 @@
 import React, {useState} from 'react';
 import {CirclePicker} from 'react-color';
-
-export default function ColorPicker(props) {
+interface ColorPickerProps {
+  color: string;
+  onUpdateColor: (color: string) => void;
+}
+const ColorPicker: React.FC<ColorPickerProps> = (props) => {
   const [isActive, setActive] = useState(false);
-
   return (
     <>
       <span className="dot" onClick={() => setActive(!isActive)} style={{color: props.color}}>
@@ -14,12 +16,11 @@ export default function ColorPicker(props) {
         <>
           <CirclePicker
             color={props.color}
-            triangle="hide"
             onChangeComplete={(color) => {
               let colorHSL = `hsl(${Math.floor(color.hsl.h)},${Math.floor(color.hsl.s * 100)}%,${Math.floor(
                 color.hsl.l * 100
               )}%)`;
-              if (colorHSL != props.color) {
+              if (colorHSL !== props.color) {
                 props.onUpdateColor(colorHSL);
               }
               setActive(false);
@@ -30,4 +31,5 @@ export default function ColorPicker(props) {
       ) : null}
     </>
   );
-}
+};
+export default ColorPicker;
--- a/src/components/Chat/ColorPicker.tsx
+++ b/src/components/Chat/ColorPicker.tsx
@@ -1,4 +1,5 @@
 import React, {useState} from 'react';
+import {useToggle} from 'react-use';
 import {CirclePicker} from 'react-color';
 import {makeStyles} from '@material-ui/core';
 interface ColorPickerProps {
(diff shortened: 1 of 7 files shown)


ewmson:
I did not do too much thinking on the discoverability here (how do users know they can change their color if they click on the dot).

Any suggestions are welcome.

src/components/Chat/ColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {CirclePicker} from 'react-color';
+interface ColorPickerProps {
+  color: string;
+  onUpdateColor: (color: string) => void;
+}
+const ColorPicker: React.FC<ColorPickerProps> = (props) => {
+  const [isActive, setActive] = useState(false);
+  return (
+    <>
+      <span className="dot" onClick={() => setActive(!isActive)} style={{color: props.color}}>
```
stevenhao, line 11: `cursor: 'pointer'` would be good here! I'm starting to standardize component styling to mui's `makeStyles`, so feel free to use that.

src/components/Chat/ColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
+import React, {useState} from 'react';
+import {CirclePicker} from 'react-color';
+interface ColorPickerProps {
+  color: string;
+  onUpdateColor: (color: string) => void;
+}
+const ColorPicker: React.FC<ColorPickerProps> = (props) => {
+  const [isActive, setActive] = useState(false);
```
stevenhao, line 8: Optional suggestion: This project has `react-use` as a dependency; you can use `const [isActive, toggleIsActive] = useToggle()` for convenience

src/pages/Game.js
```diff
@@ -208,6 +208,17 @@ export default class Game extends Component {
     return lastMessage > this.state.lastReadChat;
   }
 
+  get userColorKey() {
+    return `user_color_${window.location.href}`;
```
stevenhao, line 212: Hmm, I'm a bit curious how this part works. I _think_ user id is persistent across refreshes for any browser that has localStorage enabled, so shouldn't that guarantee that user colors are persistent as well?

stevenhao: Though I'm now thinking that if userColorKey were a constant, then that would allow users to have a consistent color across games. Eventually i want to build a proper user system w/ persistent settings, profile (name & color), etc., but in the meantime having persistent color would be cool!

src/pages/Game.js
```diff
@@ -208,6 +208,17 @@ export default class Game extends Component {
...
+  }
+
+  get userColor() {
+    let color =
```
stevenhao, line 216: The previous logic for rendering the user cursor determined the user color in a weird way (thereby denormalizing the source of truth for user color). Your refactor solves that problem by making the user cursor color derive primarily from `this.game.users` -- great! Is the localStorage part necessary?  Seems like `this.game.users[this.props.id]?.color` will be truthy in almost all cases after the initial load.

stevenhao:
This is awesome! Thanks so much for the contribution. Looks like a lot of the code for this feature already existed -- I probably had intended to build it at some point but forgot 😂 .

Re discoverability: Perhaps you can make `cursor: pointer` (see first inline comment)?

I also added some other style comments & questions; let me know if they make sense! Happy to discuss in more detail if they don't.

src/pages/Game.js
ewmson: So the problem that I was using local storage for is that we initially do not have all of the events reduced when we first load the game, so the color will be null and we will end up with a random color (and even worse changing our color to that, which means we lose whatever we would have reduced to).

To get around needing to understand and refactor the color state, I used the localstorage as a hammer so that we would at least have the color we think we should when we reload.


I can move us to storing the color directly so that the color is persisted throughout. I agree that moving to a more persistent profile system will certianly make this easier and probably would give a nice cleanup of this code.

stevenhao: Ah I see, that makes sense! I forgot that a `setColor` event was sent on client-load (what a hack 😅), that's definitely a good reason to use localStorage to workaround.

stevenhao (approved):
Awesome!

PR #148, opened 2021-01-22
[PostgresPuzzleList 1/N] Move puzzle listing to use postgres

ewmson:
- move filtering to postgres query
- move postgres to jsonb to be able to query faster (indexes coming soon too)
- revamp NewPuzzleList.tsx to change query based on new filters and fetch more with scrolling

### DIFF
--- a/server/model/puzzle.ts
+++ b/server/model/puzzle.ts
@@ -3,6 +3,7 @@ import Joi from 'joi';
 import uuid from 'uuid';
 import {pool} from './pool';
 import {PuzzleJson} from '@shared/types';
+import {ListPuzzleRequestFilters} from '../../src/shared/types';
 
 // ================ Read and Write methods used to interface with postgres ========== //
 
@@ -22,7 +23,7 @@ export async function getPuzzle(pid: string): Promise<PuzzleJson> {
 }
 
 export async function listPuzzles(
-  filter: {},
+  filter: ListPuzzleRequestFilters,
   limit: number,
   offset: number
 ): Promise<
@@ -37,11 +38,18 @@ export async function listPuzzles(
       SELECT pid, uploaded_at, content
       FROM puzzles
       WHERE is_public = true
+      AND (content->'info'->>'type') = ANY($1)
+      AND ((content -> 'info' ->> 'title') || (content->'info'->>'author')) ILIKE ALL($2)
       ORDER BY pid DESC 
-      LIMIT $1
-      OFFSET $2
+      LIMIT $3
+      OFFSET $4
     `,
-    [limit, offset]
+    [
+      mapSizeFilterForDB(filter.sizeFilter),
+      filter.nameOrTitleFilter.split(/\s/).map((s) => '%' + s + '%'),
+      limit,
+      offset,
+    ]
   );
   const puzzles = rows.map(
     (row: {pid: string; uploaded_at: string; is_public: boolean; content: PuzzleJson}) => {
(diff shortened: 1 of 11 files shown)


src/components/PuzzleList/NewPuzzleList.tsx
```diff
@@ -34,22 +28,45 @@ const NewPuzzleList: React.FC<NewPuzzleListProps> = (props) => {
...
-    setPuzzles([...puzzles, ...nextPage.puzzles]);
-    setPage(page + 1);
+    const nextPage = await fetchPuzzleList({page: currentPage, pageSize, filter: props.filter});
+    setPuzzles([...currentPuzzles, ...nextPage.puzzles]);
+    setPage(currentPage + 1);
     setLoading(false);
     setFullyLoaded(_.size(nextPage.puzzles) < pageSize);
   };
+  useEffect(() => {
+    // it is debatable if we want to blank out the current puzzles here or not,
```
stevenhao, line 55: yeah, i'm fine with this behavior

server/sql/create_puzzles.sql
```diff
@@ -10,7 +10,7 @@ IF NOT EXISTS puzzles
   uploaded_at timestamp without time zone,
 
   -- static properties of the puzzle
-  content json
+  content jsonb
```
stevenhao, line 13: This makes sense, thanks for this change. I'm on board with using jsonb since 
1) Seems more flexible and simpler than either extracting or replicating data as string columns
2) Performant enough, considering the small size of the puzzles table.

ewmson:
Decided to make this the first of a few different pull requests:

- [ support getting the correct solve count for puzzles]
- [make the puzzle actually load correctly when clicked (needs to make the puzzle model not use firebase)]
- [ add indexes to make search faster for the searches we have]

server/model/puzzle.ts
```diff
@@ -37,11 +38,18 @@ export async function listPuzzles(
       SELECT pid, uploaded_at, content
       FROM puzzles
       WHERE is_public = true
+      AND (content->'info'->>'type') = ANY($1)
+      AND ((content -> 'info' ->> 'title') || (content->'info'->>'author')) ILIKE ALL($2)
```
ewmson, line 42: This query here allows us to have the same searching logic as before (for each token is it in either the title or author?). It is a bit jank, but we can likely make an index to match this exact expression to make it quicker if desired.

ewmson: Note, I have not yet added indexes as I have not pulled in all of the puzzles locally, so this currently will make us do a full table scan.

ewmson:
For [make the puzzle actually load correctly when clicked (needs to make the puzzle model not use firebase)] , we actually load fine, I was just confused as my local pids did not match the prod ones.

stevenhao:
Just merged in https://github.com/downforacross/downforacross.com/pull/149 and https://github.com/downforacross/downforacross.com/pull/150, both prod & staging should now be up to date with firebase's puzzle repo.

- the uploaded_at field is incorrect for all imported puzzles; i just set them to today's date. This may cause some weirdness in queries in the short term, but get smoothed out as more puzzles get uploaded.

ewmson:
Cool, thanks for getting it backfilled.

ewmson:
rebased off of new master to make vercel testing easier

server/model/puzzle.ts
```diff
@@ -37,11 +38,18 @@ export async function listPuzzles(
...
       ORDER BY pid DESC
```
stevenhao, line 43: unrelated to this PR, but noticed while testing: pid is a string, so this sorts lexicographically. I'll change this sort by number (or by uploaded at) separately.

src/components/PuzzleList/PuzzleList.js
```diff
@@ -134,7 +134,17 @@ export default class PuzzleList extends PureComponent {
...
   render() {
     if (localStorage.newPuzzleListEnabled) {
-      return <NewPuzzleList />;
+      const filter = {
+        nameOrTitleFilter: this.props.search,
+        sizeFilter: this.props.sizeFilter,
+      };
+      return (
+        <NewPuzzleList
+          filter={filter}
```
stevenhao, line 143: unrelated, but should force a remount of the component when the user creates a new public puzzle

server/sql/create_puzzles.sql
stevenhao: i'll apply this in a separate PR. i think it'll look like this: ```
ALTER TABLE table_with_json
  ALTER COLUMN my_json
  SET DATA TYPE jsonb
  USING my_json::jsonb;
```

stevenhao (approved):
This is huge, thanks for helping doing this!

PR #151, opened 2021-01-24
[PostgresPuzzleList 2/N] Add an index to (name,title) for faster searching

ewmson:
We add an index to speed up the queries from 100ms to 10ms.

This does make the code slightly less readable, but we should not need to change this often so I think it is fine.

PR #153, opened 2021-01-25
[PostgresPuzzleList 3/N] Add solved stats to new puzzle list

ewmson:
Adds in the solved count for the puzzle list.

This makes a denormalized column on puzzles to get the number of solves and logs to it whenever someone does a solve.

Depends on https://github.com/downforacross/downforacross.com/pull/151 so will also have those changes until that one is merged (let me know if there is a better way to have these broken up pull requests with dependencies).


server/model/puzzle.ts
```diff
@@ -127,3 +134,32 @@ export async function addPuzzle(puzzle: PuzzleJson, isPublic = false, pid?: stri
   );
   return pid;
 }
+
+export async function recordSolve(pid: string, gid: string, timeToSolve: number) {
```
stevenhao, line 138: nice, thanks for writing this as a transaction!

server/sql/create_puzzle_solves.sql
```diff
@@ -0,0 +1,18 @@
...
+    pid text NOT NULL REFERENCES puzzles ON DELETE CASCADE,
+    gid text NOT NULL,
+    solved_time timestamp without time zone, -- the time the solve was recorded
+    time_taken_to_solve integer CHECK (time_taken_to_solve > 0), -- the duration (seconds) of how long it took to solve
+
+    constraint only_one_solve_per_puzzle_and_game UNIQUE(pid,gid)
+);
+
+ALTER TABLE public.puzzle_solves
+    OWNER to dfacadmin;
```
stevenhao, line 16: heh, thanks for copy/pasting this boiler plate even if it doesn't mean anything to your local setup. maybe we should consider using a postgres ORM to streamline & standardize the local setup for backend dev. know any good ones?

server/README.md
```diff
@@ -62,7 +65,13 @@ psql -c 'create database dfac'
 
 (`createdb` if this fails)
 
-2. Create the game_events table
+2. Create the tables
+
+```
+./create_fresh_dbs.sh
```
stevenhao, line 71: thanks for setting this up! This will definitely be really helpful as new contributors onboard to the server code.

stevenhao:
Created the table in staging and prod via
```
psql < create_puzzle_solves.sql
psql < add_times_solved_1_24_21.sql
ppsql < create_puzzle_solves.sql
ppsql < add_times_solved_1_24_21.sql
# ppsql: aliased to PGDATABASE="production" psql
```

after some changes to the ts, I was able to test this via `devbackendprod`. It's working!

image

server/README.md
```diff
@@ -15,9 +15,11 @@
 - Prod: TBD, probably downforacross.com/ws??, probably a separate process
```
stevenhao, line 15: ok i should update this readme sometime :P

stevenhao:
Wrap-up:

This PR has been fully deployed to downforacross.com.
- users must use new puzzle list (`localStorage.newPuzzleListEnabled = '1'`) to see solve count statistics
- all solves are logged to db regardless of the local flag

server/model/puzzle.ts
```diff
@@ -127,3 +137,55 @@ export async function addPuzzle(puzzle: PuzzleJson, isPublic = false, pid?: stri
...
+    [gid]
+  );
+  return count > 0;
+}
+
+export async function recordSolve(pid: string, gid: string, timeToSolve: number) {
+  const solved_time = Date.now();
+  const client = await pool.connect();
+
+  // Clients may log a solve multiple times; skip logging after the first one goes through
```
ewmson, line 161: I think this was the cause of the stalls, because we early returned without closing client.

Do we actually need this check, or can we just rely on the db to fail if we ever have a (pid,gid) pair already? (It should be super fast as it will have an index on (pid,gid)

stevenhao: yeah, you're right. my mistake 😄 .

Minor point of concern:
I'm a bit wary of leaning on the transaction rollback to ensure consistency, I'd rather think of the index as a last line of defense. My wariness comes from
- clients will possibly send a lot of recordSolve (anywhere from 1 to 10 would be typical, i think)
- semantics of exception handling typically imply that it's not a primary code path

Another proposal which I don't know how to implement: (warning -- really bad sql pseudocode)
```
1) insert (gid,pid) into puzzle_solves if not already there
2) set puzzles[pid].solves = count(pid in puzzle_solves)
```

This way, there's no error handling necessary.

stevenhao:
https://github.com/downforacross/downforacross.com/commit/56b99f3a923dde428fca562b7cf3d72f794b916f fix here

PR #156, opened 2021-01-25
Dropdown outside click

audreyli:
Kapture 2021-01-24 at 20 38 32

### DIFF
--- a/src/components/Toolbar/ActionMenu.js
+++ b/src/components/Toolbar/ActionMenu.js
@@ -1,5 +1,6 @@
 import './css/ActionMenu.css';
 import React, {Component} from 'react';
+import {findDOMNode} from 'react-dom';
 
 /*
  * Summary of ActionMenu component
@@ -24,6 +25,8 @@ import React, {Component} from 'react';
  * */
 
 export default class ActionMenu extends Component {
+  containerRef = React.createRef();
+
   constructor() {
     super();
     this.state = {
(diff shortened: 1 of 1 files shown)


src/components/Toolbar/ActionMenu.js
```diff
@@ -24,15 +25,31 @@ import React, {Component} from 'react';
...
+
   constructor() {
     super();
     this.state = {
       active: false,
     };
   }
 
+  handleMouseDown = (e) => {
+    const refNode = findDOMNode(this.containerRef.current);
```
stevenhao, line 38: What happens if you don't use `findDOMNode` here?

audreyli: it'll set it to close and then handleClick on the dropdown menu which triggers after will open it (so clicking on the button no longer closes the menu). couldn't think of a nice way to handle that case other than checking the node

stevenhao: wow that's rough. does `mouseMove={e=>e.stopPropagation()}` do anything?

audreyli: to handle this case? it shouldnt, the mouse doesnt actually move between mousedown and mouseup. also prob not good to randomly stop propagation of global mouse move event :P 

im also happy to close this in favor of using a library (material-ui) that already supports outside click to close

stevenhao: i meant attaching that to the dropdown menu itself, so that the event doesn't bubble up to window

PR #160, opened 2021-01-26
Add eslint to precommit hook

ewmson:
Adds eslint to run (with lint-staged) before every commit in the codebase (unless you specifically disable the hooks).

**NOTE: look at the first 2 commits to see the non-linting changes**

Starting off with airbnb style and then made some adjustments to make it not error out every single file.

Left some in as warnings that I thought we wanted to use for new files, but can be changed ex:
-  react/destructuring-assignment (need for const {field1,field2} = props;) instead of props.field1 is currently a warning
- no-restricted-syntax (does not like loops, and instead prefers array iterations [claims perf + minimizing side effects])
- jsx-a11y/* (left in the accessibility as warnings to have [eople change it when they see it)
- no-console (do we want to flag consolve statements as warnings?)

I made some arbitrary calls to get it all working, so am very open to changing some if we want stricter enforcement, or no warnings for the rules I mentioned earlier.

The eslintrc should be setup to work with an IDE (it does with webstorm) and can be run locally before being run on the commit hook if desired.

Errors will block the commit and warnings will not. If autofixable, then it will be done automatically.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,44 @@ module.exports = {
     },
   },
   rules: {
-    'import/extensions': 'off', // disabled as none of the current code follows this pattern.
+    // disabled as we likely do not want it
+    'no-return-await': 'off',
+    // disabled as none of the current code follows this pattern.
+    'import/extensions': 'off',
+    'react/jsx-filename-extension': 'off',
+    'react/button-has-type': 'off',
+    'react/sort-comp': 'off',
+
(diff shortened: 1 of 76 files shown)


ewmson:
need to fix why it is not deploying

stevenhao:
Thanks for setting this up! I'm on board with all the judgment calls you made, can always change them in the future if we need to

PR #161, opened 2021-02-13
Fix search

ewmson:
The new puzzle list search was always lagging behind by one character type which was not ideal.

This is fixed by making the react callback actually depend on the filters which will make the search match all of the current filters, because we will reload whenever the filter changes.

We JSON stringify as react compares with === so if we do not then it will change on every render rather than only then the filters actually change.


stevenhao (approved):
sweet, thanks for the fix! i tested on prod & https://downforacrosscom-git-fork-ewmson-fixsearch.downforacross1.vercel.app/ and indeed the 1-character lag you described was real.

PR #163, opened 2021-02-26
Show better feedback when a user uploads an invalid .puz file

JohnHKoh:
Currently, upon uploading an invalid .puz file, the upload silently fails with the unhandled exception or error showing up in the console as seen below.
- image
- image

The modals from these changes should add better feedback for the following 2 cases.

**1. When the file _is not_ in valid .puz format, this modal is shown.**
image

**2. When the file _is_ in valid .puz format, but the parsed puzzle is not a valid puzzle, this modal is shown.**
image


stevenhao:
Looks great, thanks!

PR #165, opened 2021-02-26
Fix blurry modals by overriding will-change property

JohnHKoh:
As of Chrome 53, the `will-change` CSS property will cause objects to not re-rasterize after a `transform` is applied. This causes the side effect of blurry images and texts. Read more about the issue here:
- https://googlechrome.github.io/samples/css-will-change-transform-rasterization/
- https://greensock.com/will-change/
- https://stackoverflow.com/a/54082690

The SweetAlert library uses the `will-change` property to optimize their animations, but this causes an undesirable side-effect as the text, buttons, and edges of modals appear to be blurry. This change overrides the default CSS of SweetAlert to fix this. Take note of the subtle differences below (on Chrome 88.0.4324.190).

Before:
image

After:
image

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -41,3 +41,7 @@ button {
   --main-green-1: rgba(31, 255, 61, 0.3);
   --main-green-2: rgba(31, 255, 61, 1);
 }
+
+.swal-modal {
+  will-change: unset !important;
+}


stevenhao:
nice digging. perhaps we should also open an issue in https://github.com/t4t5/sweetalert

Issue #164, opened 2021-02-26
Shift-Arrow keyboard navigation


danoa:
Make it possible to move cursor perpendicular to current orientation without changing orientation, using the same keyboard shortcuts as Across Lite. This functionality exists and is currently mapped to characters '[' and ']'. These characters are hard to press on many international keyboards, so having the additional mappings of Shift-Arrow keys would be beneficial.

stevenhao:
I haven't used Across Lite but this sounds reasonable. I can't think of any reason not to do this, PRs welcome!

danoa:
Thank you for the fix, this works great! <3

PR #167, opened 2021-02-27
Add Shift-Arrow keyboard navigation for perpendicular cursor movement

kevinjihwanlee:
This PR makes it possible to move cursor perpendicular to current orientation without changing orientation, using the same keyboard shortcuts as Across Lite (Shift-Arrow). Requested in #164

### DIFF
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -132,6 +132,12 @@ export default class GridControls extends Component {
       ']': 'forward',
     };
 
+    if (shiftKey) {
+      const isAcross = this.props.direction === 'across'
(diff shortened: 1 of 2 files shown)


src/components/Player/GridControls.js
```diff
@@ -132,6 +132,12 @@ export default class GridControls extends Component {
       ']': 'forward',
     };
 
+    if (shiftKey) {
+      const isAcross = this.props.direction === 'across'
```
stevenhao, line 136: nice, this is a clever way to get this feature working!

The way i had originally set up the key down handler didn't really scale (it passed in `shiftKey` into handleAction, which ends up being a bit clunky for cases other than tab), perhaps a refactor is in order to make it more extensible in the future.

PR #168, opened 2021-02-28
Add info modal with links to Discord and GitHub

JohnHKoh:
Added an information 'i' icon to the nav bar.
image

Clicking on the icon will pop up a modal with information about the site, including a link to the Discord and this GitHub.
image

Tested on desktop and on mobile, and this change also inadvertently fixes the search bar flex-grow/shrink that wasn't working on mobile. The 'secret' button remains to the left of the info button. The 'Log In' button, if visible, is to the right of the info button.

### DIFF
--- a/src/components/common/Nav.js
+++ b/src/components/common/Nav.js
@@ -1,9 +1,11 @@
 import './css/nav.css';
 
 import {Link} from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


stevenhao:
Love it! Thanks for helping bring more people to the discord, excited to see what that's like

PR #170, opened 2021-03-02
Use cross-env to make yarn start work for Windows

JohnHKoh:
The `{KEY}={VALUE}` syntax used to set environment variables does not work for Windows. Instead, the `set` keyword must be used as such: `set PORT=3020`. Instead of having 2 start scripts (like `yarn start` and `yarn start-windows`) to account for this difference, we can use [cross-env](https://www.npmjs.com/package/cross-env) to have a single start command without worrying about the operating system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "body-parser": "^1.19.0",
     "clsx": "^1.1.0",
     "cors": "^2.8.5",
+    "cross-env": "^7.0.3",
     "express": "^4.17.1",
     "firebase": "^7.8.1",
     "firebase-admin": "^8.9.2",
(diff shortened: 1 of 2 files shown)


stevenhao:
sounds good! i noticed cross-env is in maintenance mode, but i suppose that's not a problem for us since we're only using it in a very basic way

JohnHKoh:
> sounds good! i noticed cross-env is in maintenance mode, but i suppose that's not a problem for us since we're only using it in a very basic way

Yes, as per their note [here](https://github.com/kentcdodds/cross-env/issues/257) the package still should be updated with necessary bug fixes or critical security updates, so hopefully it shouldn't be an issue.

PR #171, opened 2021-03-04
Added modals for uploading puzzles

JohnHKoh:
Uploading a puzzle should appropriately show the right modals on successful validation, where you can choose unlisted or public. Also refreshes the puzzles list upon confirming a public upload. Also some prettier CSS with the drag and drop.

After selecting a file to upload:
image

After selecting the upload button:
image

After choosing unlisted and selecting the upload button:
image

By using the [@sweetalert/with-react](https://www.npmjs.com/package/@sweetalert/with-react) package, I was also able to remove some helper functions from the modals and allow for more extensibility and customization to the modals.

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -37,7 +37,7 @@ const actions = {
     setPuzzle(pid, puzzle);
   },
   // puzzle: { title, type, grid, clues }
-  createPuzzle: (puzzle, cbk) => {
+  createPuzzle: async (puzzle, cbk) => {
     db.ref('counters').transaction(
       (counters) => {
         const pid = ((counters && counters.pid) || 0) + 1;
--- a/src/actions.js
+++ b/src/actions.js
@@ -37,7 +37,7 @@ const actions = {
     setPuzzle(pid, puzzle);
   },
   // puzzle: { title, type, grid, clues }
-  createPuzzle: async (puzzle, cbk) => {
+  createPuzzle: (puzzle, cbk) => {
     db.ref('counters').transaction(
       (counters) => {
         const pid = ((counters && counters.pid) || 0) + 1;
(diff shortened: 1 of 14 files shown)


src/components/Upload/Upload.js
```diff
@@ -31,16 +32,17 @@ export default class Upload extends Component {
       private: !isPublic,
     };
     // store in both firebase & pg
-    actions.createPuzzle(puzzle, (pid) => {
+    await actions.createPuzzle(puzzle, (pid) => {
```
stevenhao, line 35: this await seems unnecessary, though feel free to leave it in since i'll be deleting `actions` entirely at some point since it's no longer used 😅

src/actions.js
```diff
@@ -37,7 +37,7 @@ const actions = {
     setPuzzle(pid, puzzle);
   },
   // puzzle: { title, type, grid, clues }
-  createPuzzle: (puzzle, cbk) => {
+  createPuzzle: async (puzzle, cbk) => {
```
stevenhao, line 40: i think this is unnecessary

stevenhao (approved):
tested it out on https://downforacrossc-git-fork-johnhkoh-johnhkoh-upload-modal-d-224569.vercel.app/, it works great! love the new design, thanks for helping out with this.

left inline comment on PR -- i think the async/await thing on `actions.createPuzzle` didn't end up being necessary. It's not a big deal since that code will be deleted soon anyway, though

src/actions.js
JohnHKoh: Gotcha. IntelliJ was yelling at me for ignoring the Promise returned from `db.ref('counters').transaction()` , but I guess since we're not doing anything with it anyways, I guess we don't need it. I updated the PR to remove the unnecessary async/await.

Issue #172, opened 2021-03-05
Implement Fencing (Cross Swords) Mode


carlyrobison:
This is in beta, but the remaining feature gaps are:

- [ ] Mobile friendly
- [ ] Table at the top, rows should be colored according to team
- [ ] Consider "Team Purple" and "Team Orange" not 1/2
- [ ] Team score/guesses should be tallied
- [ ] Confetti on a team winning (primarily team's color)
- [ ] Add a chat for taunting the other players!
- [ ] Can add game updates to the chat (e.g. "X joins Team Purple", "Team Purple Wins")
- [ ] Figure out what to do with clues that reference other clues
- [ ] Tutorial for how to play and what the win conditions are
- [ ] Circles in the squares

Implement these things or spawn new issues

stevenhao:
i'll work on these today:
- mobile
- color rows in table
- team score & guess tally

And hopefully will also get to these:
- "game is over" when one team reaches a majority of points
- "puzzle is complete" when puzzle is complete
- "Start game" button at the beginning, with countdown timer
- button to go back to main site
- circles

Will not get to these, but want to soon:
- ability to create fencing game from main page (without manual url hack)
- tutorial
- post-game replays (& highlights???)

Need to think more:
- clue references
- chat
- game update log
- pings
- rulesets (guess timeout, guess penalty, reveal timeout / penalty) & game modes (all start in the same corner? free for all/multiple teams? balance -- asymmetry due to long horizontal theme clues favoring left side)

stevenhao:
picking this up again now that i have some time. others are also welcome to help!

recent commits:

https://github.com/downforacross/downforacross.com/commit/3863da6341484ca4223240e5184a8ab270f2e073
https://github.com/downforacross/downforacross.com/commit/618a9b83731203935947187a53c747c88f17dcb3
https://github.com/downforacross/downforacross.com/commit/aa928f0b56c05f3b4239c5b8e566f7ca4176d3fe

earlier i also added a chat & moved around the scoreboard. will update the todos in OP to reflect state

stevenhao:
some more todos

- [ ] log when fencing game is created/user joins so that they can keep track of which ones they've seen
- [ ] prettier homepage section for variants (help wanted -- would appreciate mocks / suggestions!)
[...]

PR #173, opened 2021-03-06
Add top nav bar for fencing mode

JohnHKoh:
Just wrapped the entire fencing element around a `<Flex>` and added a `<Nav>` component.
image

Address "Banner at the top that goes back to the main site" from #172.

### DIFF
--- a/src/components/Fencing/Fencing.tsx
+++ b/src/components/Fencing/Fencing.tsx
@@ -3,6 +3,7 @@ import * as uuid from 'uuid';
 import React, {useState} from 'react';
 import {useUpdateEffect} from 'react-use';
 import {Helmet} from 'react-helmet';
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Handle line endings automatically for files detected as text
```
stevenhao, line 1: could you explain this change a bit? i'm not fully familiar with gitattributes. also perhaps it could be a separate PR

JohnHKoh: Shoot sorry, I didn’t mean to commit that. Since I’m working in Windows I needed to make sure my line endings were consistent with the rest of the repo. I’ll take it out and update this PR later.

JohnHKoh: Should be good now.

PR #180, opened 2021-03-25
Adding 's' functionality in vim mode

dangeng:
Please let me know if I'm doing this right. I have zero web dev or open source experience

Also would love to help contribute more (preferably something small and easy for a beginner), like fleshing out the vim functionality a bit more

### DIFF
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -211,7 +211,7 @@ export default class GridControls extends Component {
       } else if (key === 'i') {
         onVimInsert && onVimInsert();
       } else if (key === 's') {
-        this.handleAction('delete', shiftKey);
+        this.delete();
         onVimInsert && onVimInsert();
       }
     } else if (key === '.') {
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -210,6 +210,9 @@ export default class GridControls extends Component {
         this.selectNextClue(true);
       } else if (key === 'i') {
         onVimInsert && onVimInsert();
+      } else if (key === 's') {
+        this.handleAction('delete', shiftKey);
+        onVimInsert && onVimInsert();
       }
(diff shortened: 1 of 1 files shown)


src/components/Player/GridControls.js
```diff
@@ -210,6 +210,9 @@ export default class GridControls extends Component {
         this.selectNextClue(true);
       } else if (key === 'i') {
         onVimInsert && onVimInsert();
+      } else if (key === 's') {
+        this.handleAction('delete', shiftKey);
```
stevenhao, line 214: i think shiftKey isn't necessary for delete -- it takes no params in https://github.com/dangeng/downforacross.com/blob/master/src/components/Player/GridControls.js#L322

stevenhao: actually it would probably be cleanest to just do `this.delete()` instead of going through `handleAction`.

stevenhao (approved):
Thanks for the PR! It looks great :).

I added an inline comment in https://github.com/downforacross/downforacross.com/pull/180/files, do you think you could address it or reply to it? And then I'll merge this into the master branch!

stevenhao:
I personally don't use the vim mode since i'm used to arrow keys, but would certainly be interested in other vim features. maybe command mode could be cool? (press colon, type a number like 43A, press enter to jump there)

dangeng:
Alright, I made the change (to be honest, I was just pattern matching :) )

Oh, I've definitely caught myself typing something like "43gg" to jump to clue 43, and being confused when nothing happens. I'll try tinkering around with adding a buffer or something for the vim commands. Then you could also add things like "dw" to delete the rest of a word. Command mode seems a bit intimidating to me tbh :)

stevenhao:
Sweet, let me know if you want to chat about the design & implementation of buffer mode! there's a #contributors channel in http://downforacross.com/discord where we can discuss

PR #182, opened 2021-03-28
Fix: Can't toggle selected direction on tablet

### LINKED ISSUE #181: Tablet: Can't toggle selected direction
To reproduce:

- Load a puzzle on a real touchscreen device
- Tap any empty cell to select row
- Tap same cell a second time

Expected:

After first tap, cell is selected with "row" highlighted alongside that row's clue.
After second tap, same cell is selected with "column" highlighted alongside that column's clue.

Actual:

First tap selects the row as expected.
Second tap causes selection to "flicker" to column but then back to row.  Subsequent taps have the same behaviour.

Devices tested:

- iPad Air 2020, iPadOS 14.4.1, Safari

https://user-images.githubusercontent.com/11449340/112745602-8837fb00-8f5e-11eb-9f4f-3281483ad859.MP4

ajhyndman:
Fixes #181

After a simple touch event, modern mobile browsers fire a simulated "click" event.  I believe what's happening in this issue is that _both_ the `click` _and_ `touchend` event handlers are calling `onClick` for each tap on the screen.  This causes the state to flip from row to column mode and back.

There are more than one approach to fixing this issue:

1. At smaller screen sizes, when this app's extra "mobile" features kick in, this issue is no longer present. Activate that mode earlier, or port whatever logic changed between modes back to the tablet.
2. In  the `touchend` event handler, just before calling `onClick`,  call e.preventDefault().  This prevents the touch event from generating a simulated click event.

However, it strikes me that the "click" event generated by browsers is exactly what these touch handlers are attempting to emulate.  Unless I'm missing some context, they appear to be redundant.

This diff just removes the touch handlers from `Cell.tsx`.  I tested that this fixes the perceived issue on iPad.  I also tested on a Samsung S9 to confirm that touch events can still toggle column/row and there were no regressions to gesture-based pan and zoom.

### DIFF
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -59,6 +59,8 @@ interface Props {
  * - Grid
  * */
 export default class Cell extends React.Component<Props> {
+  private touchStart: {pageX: number; pageY: number} = {pageX: 0, pageY: 0};
+
   shouldComponentUpdate(nextProps: Props) {
     const pathsToOmit = ['cursors', 'pings', 'cellStyle'] as const;
     if (!_.isEqual(_.omit(nextProps, ...pathsToOmit), _.omit(this.props, pathsToOmit))) {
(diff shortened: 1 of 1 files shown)


stevenhao (commented):
Sorry for missing this PR until now! Appreciate all the digging and testing you did into this! I think I prefer the `e.preventDefault()` route since it's less likely to cause a regression. If I recall correctly, the start/end logic is there in case the click event doesn't fire for some reason (e.g. if the user drags their finger a bit on accident while doing the click?).

ajhyndman:
> If I recall correctly, the start/end logic is there in case the click event doesn't fire for some reason (e.g. if the user drags their finger a bit on accident while doing the click?).

My point was that browsers already implement their own heuristics for disambiguating between a short drag and a "click".  At this point, those heuristics are pretty sophisticated, so I'd suggest that re-implementing that heuristic in this codebase is probably redundant.

That said, I'm happy to update this diff with a preventDefault approach instead.

stevenhao:
Thanks for the update. I don't know enough to comment on the heuristics that most devices use for detecting `click` events and supporting short drags -- that's an area I'd need to do more research and experimentation in. For now, I'm happy with listening to `touchend` instead of `click`, and calling `e.preventDefault()` to suppress the click event.

Thanks for the contribution!

PR #186, opened 2021-04-13
Fix several issues with mobile keyboard

JohnHKoh:
On mobile, after refreshing the page or re-opening the keyboard, the first key entered would show a "$" no matter what key was pressed. This was due to the value being set to {char}$ instead of the expected ${char}, I guess due to the behavior where clicking on a textarea makes your cursor insert at the start. This changes `onfocus` to set the cursor to the end of the textarea value. See stackoverflow answer here: https://stackoverflow.com/a/10576409/4157513 (the "quirk" they mention doesn't seem to apply anymore to my version of Chrome). 

Also updated autoComplete and autoCapitalize attributes for better keyboard behavior on mobile.


src/components/Player/MobileGridControls.js
```diff
@@ -414,6 +414,7 @@ export default class MobileGridControls extends GridControls {
   }
 
   focusKeyboard() {
+    this.inputRef.current.selectionStart = this.inputRef.current.selectionEnd = this.inputRef.current.value.length;
```
stevenhao, line 417: ahhh thank you!

src/components/Player/MobileGridControls.js
```diff
@@ -492,7 +493,8 @@ export default class MobileGridControls extends GridControls {
         touchEvents: 'none',
         position: 'absolute',
       },
-      autoComplete: 'nope',
+      autoComplete: 'none',
```
stevenhao, line 496: lol, nope was actually intentional (from a comment in https://gist.github.com/niksumeiko/360164708c3b326bd1c8).
but seems like it doesn't work anyway.

stevenhao (approved):
amazing as always. thanks for looking into this!

src/components/Player/MobileGridControls.js
JohnHKoh: Actually, autoComplete: 'none' is not a real thing either... There's [autoCapitalize: 'none' ](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize) but no autoComplete: 'none'. Someone should change that to 'off'.

Issue #183, opened 2021-04-01
Shading/circle display error in 3/31/21 puzzle


frozenpandaman:
NYT display:
![Screen Shot 2021-03-31 at 18 19 09](https://user-images.githubusercontent.com/7245874/113230520-fe35ae00-924d-11eb-87d8-6b295ac59cd3.png)

Down for a Cross display:
![Screen Shot 2021-03-31 at 18 19 28](https://user-images.githubusercontent.com/7245874/113230537-08f04300-924e-11eb-8972-3ea2b8aab253.png)

Plasser1:
This is happening in the the 4/18/21 NYT Puzzle as well
https://downforacross.com/beta/play/13422
Correct circle placements: https://crosswordfiend.com/2021/04/17/sunday-april-18-2021/#ny

stevenhao:
this must be a bug in the puz parser. i wonder if nyt changed something about their encoding.

frozenpandaman:
@stevenhao found another likely .puz parser bug, in 4/21/2020:
![image](https://user-images.githubusercontent.com/7245874/115648300-0e372f80-a2da-11eb-8f0c-fe575e2366a0.png)
![image](https://user-images.githubusercontent.com/7245874/115648320-142d1080-a2da-11eb-9b8b-5d24e39324c6.png)

Should be "R2D2" and "C3PO." Seems to think numbers should be Ts.

cameronfreer:
See also #122

JohnHKoh:
Does anyone have a .puz file they can share the demonstrates the incorrect behavior? That could help in getting to the bottom of this.

jackofallspades98:
github doesn't allow me to attach .puz files in comments, but here's a file that demonstrates the circle issue when uploaded to DfA
https://drive.google.com/file/d/1P_DehmxORtNi2F4rT8NbQCtE3xgIRfaw/view?usp=sharing

stevenhao:
Thanks so much @JohnHKoh  for the fix!

stevenhao:
Turns out the bug was not in the parser, just in the frontend (did not support rectangular grids due to a bug that did a computation based on the number of rows instead of number of columns)

Issue #122, opened 2020-10-11
Circles are in the wrong spot for some puzzles


cameronfreer:
On several, but not all, circle puzzles (see example below), the circles are in the wrong spot -- sometimes elsewhere within the same word, other times farther away from the correct location. As far as I can tell, they are generally in the same wrong spot for all copies of the same puzzle, and have the correct number of circles, just wrong locations. Maybe there's some issue with the .puz parser?

Example:
NY Times, Sun, Aug 05, 2018 GHOSTED
Daily Puzzle | By Alison Ohringer and Erik Agard / Will Shortz
https://downforacross.com/beta/play/9302
https://downforacross.com/beta/play/6313
https://downforacross.com/beta/play/2992
![image](https://user-images.githubusercontent.com/2009342/95688625-80b3ae00-0bd9-11eb-8e69-41a1588992f1.png)

Correct location (from [xwordinfo](https://www.xwordinfo.com/Crossword?date=8/5/2018))
![image](https://user-images.githubusercontent.com/2009342/95688656-a6d94e00-0bd9-11eb-9463-ad6e913a8fa3.png)

stevenhao:
hmm, interesting. agree this is probably a bug with the puz parser.

cameronfreer:
Here's another example:

NY Times, Tuesday, September 5, 2017
Daily Puzzle | By Michelle Kenney and Jeff Chen / Will Shortz
https://downforacross.com/beta/play/10144
![image](https://user-images.githubusercontent.com/2009342/101291295-b0e08d00-37d5-11eb-94fc-ededaa422c00.png)

Correct location (from [xwordinfo](https://www.xwordinfo.com/Crossword?date=9/5/2017))
![image](https://user-images.githubusercontent.com/2009342/101291312-d5d50000-37d5-11eb-8853-eabf5b209e03.png)

PR #187, opened 2021-04-24
Fix incorrect circle markups in some puzzles

JohnHKoh:
toCellIndex requires number of columns, but the number of rows was being passed in, leading to the circles being the wrong place. Passing in grid[0].length fixes this. Addresses #122 and #183.


stevenhao (approved):
ahhhh thank you!

frozenpandaman:
cool, thanks!

PR #205, opened 2021-09-03
Add color picker to pencil tool

JohnHKoh:
Adds a color picker to the pencil tool.
image

### DIFF
--- a/src/components/Toolbar/index.js
+++ b/src/components/Toolbar/index.js
@@ -25,9 +25,19 @@ export default class Toolbar extends Component {
 
   handlePencilColorPickerClick = (e) => {
     e.stopPropagation();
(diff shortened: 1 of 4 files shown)


src/components/Toolbar/index.js
```diff
@@ -23,6 +23,28 @@ export default class Toolbar extends Component {
...
+          return hex + hex;
+        })
+        .join('');
+    }
+    this.pencilColorPicker.value = '#' + hexColor;
+    this.pencilColorPicker.click();
+  };
+
+  handlePencilColorPickerChange = (e) => {
+    document.documentElement.style.setProperty('--pencil-color', e.target.value);
```
stevenhao, line 45: i'm fine w/ storing this global state on dom instead of in react, could one day persist to browser storage too

stevenhao (approved):
sweet, looks great!

PR #206, opened 2021-09-03
Add keyboard shortcuts for check/reveal

JohnHKoh:
Adds keyboard shortcuts for check/reveal square, word, and puzzle. No confirmation dialog.

Alt + S, W, or P - Check square, word, or puzzle
Alt + Shift + S, W, or P - Reveal square, word, or puzzle

image


stevenhao:
Nice. S and W are also conveniently close to each other & located on left hand (so users can use right hand for mouse i guess?)

Issue #203, opened 2021-08-21
Empty root div created for puzzle with no clues?


jwarley:
https://downforacross.com/beta/play/15819

"Can you solve this one? Tricky tricky...."

frozenpandaman:
tricky tricky indeed

JohnHKoh:
Fixed with #208, though ideally, we should not even allow such puzzles to be uploaded in the first place, unless unlisted.

jwarley:
![image](https://user-images.githubusercontent.com/12537221/133025386-1981fa7e-2e72-408c-80c3-70ebdc09aa57.png)

Victory at last.

Issue #199, opened 2021-07-21
Defaults to "0A" if there's no 1 across


frozenpandaman:
Not sure if this is technically a bug, but the site by default (on page load) loads a null clue "0A" if there's a nonstandard crossword without a 1 across, e.g. https://downforacross.com/beta/game/1083156-trag

![image](https://user-images.githubusercontent.com/7245874/126454471-50ee2b05-6bcf-4c25-95d8-f05765bd6a87.png)

In this case I'd expect either 1 down or 4 across to be loaded, but the current behavior isn't too bad IMO, just nothing gets loaded by default.

JohnHKoh:
Fixed with #208. 4 across is selected in this case.

PR #208, opened 2021-09-04
Update logic for setting initial selected squares

JohnHKoh:
Updates logic for setting the initially selected squares on puzzle load.

Selects first available 'across' clues, and then first available 'down' clues. Defaults to selecting {0,0} if none found. For example, if a puzzle does not have 1A, but has 4A (see https://downforacross.com/beta/game/1083156-trag), the 4A squares will be selected on page load.

Handles edge cases where puzzles:
-Don't have 1A clue
-Don't have any across clues
-Don't have any clues (though we should not allow such puzzles to be uploaded in the first place)

Should address #199 and #203.

### DIFF
--- a/src/components/Player/Player.js
+++ b/src/components/Player/Player.js
@@ -42,12 +42,13 @@ const PING_TIMEOUT = 10000;
 export default class Player extends Component {
   constructor(props) {
     super(props);
+    const selected =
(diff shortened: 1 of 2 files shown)


stevenhao (approved):
thanks for fixing this!

Issue #175, opened 2021-03-07
Add timestamps to chat


tcdachille:
I'd like to see timestamps added to chat, which would be especially useful when collaborating on a puzzle over an extended period of time

stevenhao:
Good idea! The data should exist already, so it's just a matter of displaying it in the frontend.

tcdachille:
Cool! Can I actually grab this task? I'd love to start contributing but don't have a ton of web dev experience

stevenhao:
That'd be awesome! Are you in the discord? I or another contributor can help you get started if you'd like.

tcdachille:
Sweet! Help getting started would be great. And yep! I joined the discord right when I was opening this issue.

kevinl120:
Is this still being worked on? If not I'd like to take a stab at it!

stevenhao:
Go for it @kevinl120

bdenney:
Can we close this issue? It appears to have been fixed!

stevenhao:
yup thanks again @kevinl120

PR #221, opened 2021-12-13
Add timestamps to chat messages

kevinl120:
Add timestamps to chat messages.

Screen Shot 2021-12-13 at 1 24 41 PM

Addresses #175.

### DIFF
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -210,6 +210,14 @@ export default class Chat extends Component {
     );
   }
 
+  renderMessageTimestamp(timestamp) {
+    return (
+      <span className="chat--message--timestamp">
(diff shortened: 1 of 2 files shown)


stevenhao:
design: maybe we can make this float to the far right, and instead of bracket delimeters, use gray/italic text styling (can probably draw design inspiration for this from another site...)

kevinl120:
Updated the PR. Timestamps are aligned right in gray, with a darker tint on hover or selection. 

Screen Shot 2021-12-13 at 6 04 57 PM

This change also makes wrapped messages align with the colon. I think it looks a little better this way, but if it's undesired I can adjust it back.

stevenhao:
awesome, thank you!

PR #224, opened 2022-02-01
Mobile Control Enhancement

bdenney:
Bound the comma key on a mobile keyboard to replicate the behavior of tab on web. Moving between clues is cumbersome without it (in my opinion of course 😄 ).

I'm completely unsure if this is the right way to go about suggesting a change like this so please let me know!


stevenhao (approved):
great idea!

stevenhao:
maybe enter would be good too?

stevenhao:
testing on https://appetize.io/demo?device=nexus7&scale=50&osVersion=9.0 nexus 
https://user-images.githubusercontent.com/1988030/152035129-4915c5d2-e1e1-471d-bb08-3ebbcef1fbd5.mov

stevenhao:
merging this for now. we can look into adding hotkey for enter in the future.

Issue #220, opened 2021-12-09
Markdown/html parsing issue in clues


frozenpandaman:
Not sure if the NYT has changed their format or something, where the underlying representation of puzzles is written using Markdown styling rather than HTML, but here is a clue on the NYT site:
<img width="251" alt="Screen Shot 2021-12-08 at 22 44 58" src="https://user-images.githubusercontent.com/7245874/145370861-2c6b9ff7-4c72-412a-af26-89c97ac2ac4d.png">

And how it shows up on downforacross:
<img width="189" alt="Screen Shot 2021-12-08 at 22 45 29" src="https://user-images.githubusercontent.com/7245874/145370881-05bab5a2-6c29-4dd1-ac4e-4e9b4c2af2b4.png">

bdenney:
@frozenpandaman which date was this clue from?

frozenpandaman:
@bdenney The date I posted it, December 9, 2021 (I just googled it to double-check 😉)

bdenney:
This is odd, but it appears to be fixed on latest main?

<img width="212" alt="Screen Shot 2022-02-01 at 2 02 02 PM" src="https://user-images.githubusercontent.com/1355090/152033633-9dc585af-d117-4682-a519-e251ae08498d.png">

stevenhao:
thanks @bdenney ! 

there are different versions of the same puzzle -- usually because the puz files were prepared by different software. i think it's pretty safe to detect html escape sequences like &amp; and &lquot; and such. 

autodetecting and parsing markdown might be a bit more tricrky; i could see it backfiring in other cases (i.e. where markdown syntax was not intended but a markdown parser picks it up inadvertently). worth a shot though

frozenpandaman:
Cool, thanks for the info! And glad it's fixed (huh! sorry, don't know the exact puzzle/game URL)

I'll close this for now then but @bdenney, of course do feel free to open a new issue for other parsing stuff. :D

stevenhao:
i think bdenney's proposal to fix the html parsing is good; i updated this issue to include html in scope too

frozenpandaman:
also probably related: #243 

cc @bdenney

PR #225, opened 2022-02-03
Decode HTML in Clues

bdenney:
Partially addresses #220 . Decoding HTML encoded clue values. Very unsure if this is the _best_ way to fix this, but it does appear to work 😄 

The puzzle provided in #220 does have a few examples of this but it doesn't appear I can upload that puz file to the issue 😢

### DIFF
--- a/src/components/Player/ClueText.js
+++ b/src/components/Player/ClueText.js
@@ -1,7 +1,14 @@
 import React from 'react';
 
+function decodeHtml(htmlText) {
+  let text = document.createElement("textarea");
+  text.innerHTML = htmlText;
+  return text.value;
+}
+
 export default ({text = ''}) => {
   const parts = [];
+
   while (text.length > 0) {
     const s = text.indexOf('<i>');
     const e = text.indexOf('</i>');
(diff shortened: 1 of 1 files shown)


src/components/Player/ClueText.js
```diff
@@ -1,5 +1,11 @@
 import React from 'react';
 
+function decodeHtml(htmlText) {
```
stevenhao, line 3: any reason not to use this? https://lodash.com/docs/4.17.15#unescape

bdenney: I tried that and it wouldn't unescape the bulk of anything in the test puzzle mentioned in #220. I found this explanation: https://stackoverflow.com/questions/36352155/unescape-function-from-underscore-js-not-working

bdenney: https://github.com/mathiasbynens/he would also be an option, but not sure whether bringing a library in for this is appropriate. (As in genuinely don't know, pretty new to JS development 😄 )

stevenhao: gotcha! i think importing he or doing it as you did are both reasonable! thanks for the explanation

PR #235, opened 2022-04-29
Implement Home and End buttons

Denton-L:
Implement these buttons so that they go to the beginning or end of a
clue, respectively.

### DIFF
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -20,6 +20,8 @@ export default class GridControls extends Component {
     right: this.setDirectionWithCallback('across', this.moveSelectedBy(0, 1).bind(this)).bind(this),
     forward: this.moveSelectedUsingDirection(1).bind(this),
     backward: this.moveSelectedUsingDirection(-1).bind(this),
+    home: this.moveToEdge(true).bind(this),
+    end: this.moveToEdge(false).bind(this),
     backspace: this.backspace.bind(this),
     delete: this.delete.bind(this),
     tab: this.selectNextClue.bind(this),
@@ -94,6 +96,17 @@ export default class GridControls extends Component {
     };
   }
 
+  moveToEdge(start) {
+    return () => {
+      const {selected, direction} = this.props;
+      let {r, c} = selected;
+      ({r, c} = this.grid.getEdge(r, c, direction, start));
+      if (this.grid.isInBounds(r, c)) {
+        this.setSelected({r, c});
+      }
+    };
+  }
+
   setDirectionWithCallback(direction, cbk) {
     return () => {
       if (this.props.direction !== direction) {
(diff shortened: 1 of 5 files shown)


src/components/Player/GridControls.js
```diff
@@ -20,6 +20,8 @@ export default class GridControls extends Component {
     right: this.setDirectionWithCallback('across', this.moveSelectedBy(0, 1).bind(this)).bind(this),
     forward: this.moveSelectedUsingDirection(1).bind(this),
     backward: this.moveSelectedUsingDirection(-1).bind(this),
+    home: this.moveToEdge(true),
```
stevenhao, line 23: hmm, this reminds me a bit of ctrl a / ctrl e / ctrl u. that might be interesting too

src/components/Toolbar/index.js
```diff
@@ -263,6 +263,10 @@ export default class Toolbar extends Component {
                 <td>Space bar</td>
                 <td>Flip orientation between down/across</td>
               </tr>
+              <tr>
+                <td><code>Home</code> OR <code>End</code></td>
```
stevenhao, line 267: at some point we may want to create a separate section for advanced shortcuts like Home/End/[/]/Tab, so that new users aren't overwhelmed with the size of the list

src/components/Player/GridControls.js
Denton-L: I'll leave this as is for now but if someone else is interested in adding this feature, it shouldn't be too difficult to do.

src/components/Toolbar/index.js
Denton-L: Good idea! While we're doing that we have some instructions that render wrong. For example,

```html
                  <code>[</code> and
                  <code>]</code> OR
                  <code>Shift</code> with arrow keys
```

ends up rendering as `[ and] ORShift with arrow keys` so we should do something like

```html
                  <code>[</code> and <code>]</code> OR <code>Shift</code> with arrow keys
```

while making that change.

stevenhao: Nice find and agreed!

Would you be able to include building the advanced section as part of this PR, or would you rather wait until that is done? I'm not entirely comfortable adding more keyboard shortcuts until we can be confident that they won't confuse new users, so that does mean it is somewhat blocking for your feature.

stevenhao:
image

let me check on the vercel settings to see if i can make build outputs public. i may just move to netlify

Denton-L:
Thanks, I tried to implement both suggestions in two PRs. I based this change on top of the commit that should separate the help lists. However, the build seems to still be failing, could you please send me the error output?

stevenhao:
thanks so much for this! i'll take a look at this in a bit.

Denton-L:
Thanks! Hopefully it should build now :)

Denton-L:
@stevenhao Sorry for the bump but could you please take a look at this when you're not busy?

stevenhao:
Hey, this looks good! Thanks for building this :)

PR #238, opened 2022-05-30
Separate "Keyboard Shortcuts"

Denton-L:
Instead of having a giant list of Keyboard Shortcuts, separate the list
into Basic and Advanced lists.

While we're at it, fix some instructions that render wrong. For example,

	<code>[</code> and
	<code>]</code> OR
	<code>Shift</code> with arrow keys

ends up rendering as

	[ and] ORShift with arrow keys

so fix the rendering by combining the lines together so that the space
isn't lost.


stevenhao:
Thanks @Denton-L! Just getting around to this, looks amazing!

before:
image


after:
image

PR #253, opened 2022-12-21
Tweak margin around chat system messages

JohnHKoh:
Changed `margin-bottom: 10px;` to a `margin: 5px;` to add a little more space around system chat messages. I didn't like how close the message was to the horizontal bar below the player names. No one really asked for this, but I think it looks nicer 🤠 If there are more system messages to be added in the future, this will come in handy.

Before:
image

After:
image


stevenhao:
lgtm!

PR #254, opened 2022-12-21
Remove border-right on toolbar

JohnHKoh:
Fixes double border caused by the toolbar and chat.

Before:
image

After:
image


stevenhao:
nice, good catch

PR #255, opened 2022-12-29
Make backend development more windows-friendly

JohnHKoh:
Adding `cross-env` to the npm scripts sections for `devfrontend` and `devbackend`. I presume these will need to be added for the other scripts as well, but just adding these in for now.

Also making a note about DirEnv since it does not come pre-packaged with Windows, and added a few more env variables that might cause issues if not automatically set. For reference, the following error is returned if the PGUSER/PGPASSWORD is not set.

```
TypeError [ERR_INVALID_ARG_TYPE]: The "key" argument must be of type string or an instance of Buffer, TypedArray, DataView, or KeyObject. Received null
```

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -1,1 +1,3 @@
 export PGDATABASE=dfac
+export PGUSER=postgres
+export PGPASSWORD=password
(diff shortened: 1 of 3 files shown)


stevenhao:
great, thanks for making this cleaner!

PR #261, opened 2023-01-14
Change dark mode colors

pkmnnerd:
Screenshot 2023-01-13 11 34 09 PM

Make some changes to dark mode colors following some guidelines from [https://m2.material.io/design/color/dark-theme.html](https://m2.material.io/design/color/dark-theme.html):
- Changed all primary text to white with 87% opacity
- Changed background to #121212
- Use 5% overlay for "elevated surfaces" and 12% for "menu surfaces" and the grid
- Use 47% opacity instead of the recommended 38% opacity for "disabled" text to maintain a 4.5:1 contrast ratio

Other color changes include
- Make black squares black and normal grid squares lighter.
- Make the selected square the users color
- Add a darker version of the referenced color
- Make highlighted row green on completion
- Modify some other colors to look better (in my opinion)

Also fixed an issue with the dark mode selection changing the height of the nav bar in mobile (when it overflows).

I'm sure this doesn't cover every possible change for a proper dark mode, but it should be a decent improvement over the current dark mode. I'm happy to modify any of these changes depending on what people think about it.

### DIFF
--- a/src/dark.css
+++ b/src/dark.css
@@ -4,6 +4,9 @@
   font-family: courier;
   opacity: 0;
   cursor: pointer;
+  text-align: right;
+  margin-top: -10px; /* Ignore padding from parent so dark mode setting doesn't expand Nav bar */
+  margin-bottom: -10px; /* Can remove these if there's a more compact dark mode switching UI */
 }
 
 .molester-moon:hover {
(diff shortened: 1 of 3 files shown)


stevenhao:
missed this somehow, my b. do you have any suggestions on launch plan? (e.g. ask in discord for testers, or just ship it now & wait for feedback)

src/dark.css
```diff
@@ -116,29 +173,45 @@
...
+.dark .entry {
+  background-color: var(--background-1);
+  border-color: rgba(255,255,255,0.3);
 }
 
 :root {
-  --dark-blue: rgb(36, 79, 115);
+  --dark-blue-1: rgb(36, 79, 115);
+  --dark-blue-2: #183a60;
+  --primary-text: rgba(255, 255, 255, 0.87);
```
stevenhao, line 213: can we prefix these with `dark`?

pkmnnerd: done

pkmnnerd:
> do you have any suggestions on launch plan? (e.g. ask in discord for testers, or just ship it now & wait for feedback)

I think it would be fine to get some feedback before pushing, but it's impossible to make everyone happy. A survey of yes or no would make sense (as opposed to asking for changes). Just merging it and getting feedback is alright too, especially if not that many people use dark mode right now.

stevenhao:
yeah i guess dark mode is pretty "unofficial" so let's just ship this and see what ppl think :)

stevenhao:
thanks for your help here @pkmnnerd! i think it looks much nicer :)

PR #264, opened 2023-02-11
Vim Mode: Enable with keyboard shortcut

bkimmig:
# What

Add "extras" menu drop down with "vim mode".

Selecting the drop down gives you the opportunity to "Enable Vim Mode"

enable-vim-mode

Once enabled the dropdown gives you the choice to "Disable Vim Mode"

disable-vim-mode

<details>
    <summary> First Proposal - Click to Expand </summary>

**See discussion below for reasons we did not do this**

Add a keyboard shortcut to enable/disable the existing `vimMode`.

- Used `:` as the hot key
  + felt right in the framing of vim
  + requires a user to press two keys to enable - less likely to happen by accident
- Added status to the info tool bar so you can see the state
- Can toggle on/off regardless of which mode you are in

## Vim Disabled
vim-disabled

## Vim Enabled
vim-enabled
</details>

### DIFF
--- a/src/components/Game/Game.js
+++ b/src/components/Game/Game.js
@@ -194,7 +194,6 @@ export default class Game extends Component {
   };
 
   handlePressPeriod = this.handleTogglePencil;
-  handlePressColon = this.handleToggleVimMode;
 
   handlePressEnter = () => {
     this.props.onUnfocus();
@@ -281,7 +280,6 @@ export default class Game extends Component {
         addPing={this.handleAddPing}
         onPressEnter={this.handlePressEnter}
         onPressPeriod={this.handlePressPeriod}
-        onPressColon={this.handlePressColon}
         listMode={this.state.listMode}
         vimMode={this.state.vimMode}
         vimInsert={this.state.vimInsert}
@@ -326,6 +324,7 @@ export default class Game extends Component {
         onReset={this.handleReset}
         onKeybind={this.handleKeybind}
         onTogglePencil={this.handleTogglePencil}
+        onToggleVimMode={this.handleToggleVimMode}
         onToggleAutocheck={this.handleToggleAutocheck}
         onToggleListView={this.handleToggleListView}
         onToggleChat={this.handleToggleChat}
--- a/src/components/Game/Game.js
+++ b/src/components/Game/Game.js
@@ -149,6 +149,12 @@ export default class Game extends Component {
     });
   };
 
+  handleToggleVimMode = () => {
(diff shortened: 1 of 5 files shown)


stevenhao:
hmm, i feel like it could still be pressed on accident. also, : doesn't seem like the right toggle since it's more of a thing you do once you're in vim already?

i don't have any good suggestions on what would be a good hotkey though... maybe a button under a menu somewhere.

bkimmig:
>i feel like it could still be pressed on accident.

Yeah, you are totally correct here after I play tested a bit more.

>i don't have any good suggestions on what would be a good hotkey though... maybe a button under a menu somewhere.

I figured since it was removed from the menu it shouldn't live there, but if you'd be ok w/ it in the menu I'd be happy to add it somewhere in there.

Alternatively we could make the hot key combo more complex so you need like 3 keys or more to avoid accidental toggling? `ctrl-alt-V` or something in that sort of space?

However, if you are more of the mind that "this isn't a feature we want" that is fine and I can move on :).

stevenhao:
> However, if you are more of the mind that "this isn't a feature we want" that is fine and I can move on :).

Personally, I like vim mode enough that i want it to exist and be somewhat discoverable (not necessarily on their first experience -- maybe on a user's 10-50th game or something). i don't personally use it (though i do use vim for programming) but i know there are users (e.g. in the discord) who do use it.

I moved it out of the "Modes Dropdown Menu" as part of a redesign of the toolbar (moving more towards icons & making color attribution its own menu button). Perhaps we should add a button to the toolbar that, when clicked, drops down a menu of "Other stuff", where we hide all the uncommonly used features :)

bkimmig:
>For vim mode, and other "niche features" Perhaps we should add a button to the toolbar that, when clicked, drops down a menu of "Other stuff", where we hide all the uncommonly used features :)

Sounds good to me. I'll close out this request and start that as a feature if that is ok?

stevenhao:
sure, or repurpose this PR to keep discussion in one place

bkimmig:
@stevenhao I've updated the PR description (collapsed the old one, but kept it). What do you think of this solution?

stevenhao:
OK! I may want to change this to an icon in the future, for the sake of saving toolbar space. Down to merge this as-is for now. Thanks for working on this!

stevenhao:
I might also add Color Attribution Mode, List View Mode, Autocheck here too to improve their discoverability. They already have icons in the toolbar, but it's not always obvious how to use them at first glance.

Issue #260, opened 2023-01-13
Win Celebration on Mobile


lukevp:
We use your site/app on mobile quite a bit, and it took us some time to realize that the only indication that we've solved the puzzle on mobile is the appearance of parenthesis and pausing of solve time, plus the highlight of green on the puzzle.  On desktop, it's really obvious you've won because of the confetti burst introduced in #96 , which I only discovered many weeks later.

That PR specifically didn't show confetti on mobile because of "performance issues".  Is that a real concern, or was that just an assumption?  I can't imagine the performance is really a big deal for a couple of seconds, if anything.  Granted, I have a nice phone (iPhone 13 Pro Max) but I tried several js confetti implementations and saw no slowness on any of them.  Since the animation only plays for a few seconds, I don't see this being a major issue.

Would you be open to adding confetti support to mobile? It should be straightforward to remove the mobile check from the existing PR, and I'd be happy to do so if you're open to the change.

stevenhao:
no oblections from me! @cleercode any thoughts?

frozenpandaman:
+1

frozenpandaman:
@lukevp @stevenhao Seems this issue can be closed now!

raghuvv:
This might be useful if someone is looking to improve the performance of "confetti" celebration across platforms:

https://github.com/catdad/canvas-confetti

PR #265, opened 2023-02-17
Turn on confetti for mobile solves

combscCode:
# What
Removes check for mobile when deciding to display confetti. Previously it wouldn't display confetti when on mobile, now always displays it.

# Why
Per issue #260 there seems to be consensus that performance issues are no longer a concern for mobile users. And confetti is fun! So let's turn it on : )


stevenhao:
sure let's do it

lukevp:
We noticed the confetti yesterday! Thanks for listening to the feedback, we’ve really enjoyed it!

Issue #246, opened 2022-08-09
.ipuz support please! :)


frozenpandaman:
Hi! Avid user of the site (and bug-reporter, haha). Just wanted to say that would be really wonderful to see support for uploading [.ipuz](http://www.ipuz.org/) files – an open, free, extensible format for crossword puzzles. It's underlyingly JSON and has detailed documentation available at that link, and there's already some [good code](https://github.com/crosswordnexus/crossword-tools/blob/main/jscrossword/ipuz_read_write.js) out there for parsing it in JS. Not sure how/where in the codebase you're parsing the .puz files currently, but I can look into helping out with adding this feature or making a PR myself if you can point me in the right direction!

stevenhao:
Cool, happy to review a PR!

this is the puz converter source, pretty much entirely home grown / adapted from puz converters in other languages :)
https://github.com/downforacross/downforacross.com/blob/master/src/lib/converter/PUZtoJSON.js#L179

invoked here https://github.com/downforacross/downforacross.com/blob/11495fb8bef02a680f2169bd90104c6bdbd721e8/src/components/Upload/FileUploader.js

frozenpandaman:
@stevenhao Oh, this is awesome, since it's just going from JSON to JSON haha. Should be very doable. Will get some time to poke into it soon. Thanks!

Hmm, I might even suggest you switch over to representing the puzzles in ipuz format under the hood… obviously that's a larger decision that you'd need to get on-board with, but looking at `FileUploader.js`, it's already like 90% there!

FYI, the basic structure of an ipuz file is just this:
```json
{
	"version": "http://ipuz.org/v1",
	"kind": [ "http://ipuz.org/crossword#1" ],
	"dimensions": { "width": 15, "height": 15 }, // or whatever
	"author": "...",
	"title": "...",
	"notes": "...",
	"copyright": "...", // all of these fields are optional
	"puzzle": [
		[1, 2, 3, 4, "#", "#", "#", 5, 6, 7, 8, 9, 10, 11, 12],
		// etc...
		// shaded square: {"cell": num, "style": {"color": "#dcdcdc"}}
		// circle in square: {"cell": num, "style": {"shapebg": "circle"}}
	],
	"solution": [
		["T", "E", "S", "T", "#", "#", "#", "REBUS", "E", "X", "A", "M", "P", "L", "E"],
		// etc...
	],
	"clues": {
		"Across": [
			[1, "blah"],
			// etc...
		],
		"Down": [
[...]

PR #267, opened 2023-04-09
Add in iPUZ support

scouttyg:
This PR adds fairly robust support in for iPUZ files, along with some utility classes to hopefully detect other crossword filetype support in the future (e.g. `jpz`).

Will add more comments inline in the PR soon!

Relevant background on iPUZ: http://www.ipuz.org/

See also #246 

## Screenshots / Demo:
ipuz-support

### DIFF
--- a/src/components/Upload/FileUploader.js
+++ b/src/components/Upload/FileUploader.js
@@ -7,6 +7,41 @@ import swal from '@sweetalert/with-react';
 
 import {hasShape} from '../../lib/jsUtils';
 import PUZtoJSON from '../../lib/converter/PUZtoJSON';
+import iPUZtoJSON from '../../lib/converter/iPUZtoJSON';
+import fileTypeGuesser from '../../lib/fileTypeGuesser';
+
+class InvalidFileTypeError extends Error {
+  constructor(fileType) {
+    const title = `Invalid .${fileType} file`;
+    super(title);
+    this.errorType = 'InvalidFileTypeError';
+    this.errorTitle = title;
+    this.errorText = `The uploaded file is not a valid .${fileType} file.`;
+    this.errorIcon = 'warning';
+  }
+}
+
+class UnknownFileTypeError extends Error {
+  constructor(fileType) {
+    const title = `Unknown file type: .${fileType}`;
+    super(title);
+    this.errorType = 'UnknownFileTypeError';
+    this.errorTitle = title;
+    this.errorText = 'The uploaded file could not be recognized';
+    this.errorIcon = 'warning';
+  }
+}
+
+class UnsupportedFileTypeError extends Error {
+  constructor(fileType) {
+    const title = `Unsupported file type: .${fileType}`;
+    super(title);
+    this.errorType = 'UnsupportedFileTypeError';
+    this.errorTitle = title;
+    this.errorText = 'The uploaded file is not currently supported';
+    this.errorIcon = 'warning';
+  }
+}
 
 export default class FileUploader extends Component {
   validPuzzle(puzzle) {
@@ -23,7 +58,6 @@ export default class FileUploader extends Component {
         down: {},
       },
     };
-
     return hasShape(puzzle, shape);
   }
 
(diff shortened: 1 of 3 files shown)


src/components/Upload/FileUploader.js
```diff
@@ -52,23 +86,63 @@ export default class FileUploader extends Component {
     return result;
   }
 
+  convertIPUZ(readerResult) {
+    const {grid, info, circles, shades, across, down} = iPUZtoJSON(readerResult);
```
scouttyg, line 90: I tried to follow the pattern here that was done for converting to `puz`, so that way things would be fairly easy to follow.

src/lib/fileTypeGuesser.js
```diff
@@ -0,0 +1,48 @@
...
+    if (e.message.toLowerCase().includes('is not valid json')) {
+      isJSON = false;
+    }
+  }
+
+  if (isJSON && parsedJson?.version && parsedJson?.version.startsWith('http://ipuz.org')) {
+    return 'ipuz';
+  }
+
+  const puzMagicHeader = '4143524f535326444f574e00';
```
scouttyg, line 17: A `.puz` file should have this magic header as per the docs (See: https://gist.github.com/sliminality/dab21fa834eae0a70193c7cd69c356d5#header)

src/lib/fileTypeGuesser.js
```diff
@@ -0,0 +1,48 @@
...
+  const otherMagicHeaders = {
+    '504b0304': {
+      mime: 'application/zip',
+      fileTypeGuess: 'jpz',
+    },
+    '3c3f786d': {
+      mime: 'text/xml',
+      fileTypeGuess: 'jpz',
+    },
+  };
```
scouttyg, line 28: `jpz` files can either be `zipped` or uncompressed `xml` files.

src/components/Upload/FileUploader.js
```diff
@@ -52,23 +86,63 @@ export default class FileUploader extends Component {
...
+        if (e?.errorText) defaultText = e.errorText;
+        if (e?.errorIcon) defaultIcon = e.errorIcon;
+
         swal({
-          title: `Invalid .puz file`,
-          text: `The uploaded file is not a valid .puz file.`,
-          icon: 'warning',
+          title: defaultTitle,
+          text: defaultText,
+          icon: defaultIcon,
```
scouttyg, line 145: Abstracted the error handling out a bit to show some of the messages of the errors:
error-handling

src/lib/converter/iPUZtoJSON.js
```diff
@@ -0,0 +1,61 @@
...
+    row.map((cell) => (cell === null || cell === '#' ? '.' : cell))
+  );
+  const info = {
+    type: grid.length > 10 ? 'Daily Puzzle' : 'Mini Puzzle',
+    title: jsonFromReader.title || '',
+    author: jsonFromReader.author || '',
+    description: jsonFromReader.description || '',
+  };
+  const circles = [];
+  const shades = [];
```
scouttyg, line 32: As a side note I don't think shades actually ever worked for normal `puz` files currently -- I believe it will always read as circles.

For `ipuz`, I think we can read `shades` in if there is a `cell.style.color`, but I couldn't find any good examples of this when I used tools like Crossword Scraper to download `ipuz` files, it always just set the `cell.style.shapebg`

stevenhao: i see, that's interesting. 
re: `shades` i don't remember exactly, but i think there were some non-standard puz files that did set this field. in any case those are probably irrelevant so i agree with your comment.

re: ipuz shades, sounds good to me for now. feel free to make a follow-up PR if you want to update this logic for ipuz!

src/components/Upload/FileUploader.js
stevenhao: great!

src/components/Upload/FileUploader.js
stevenhao: i'd usually do something like this, or even just inline it into the call to `swal`, but fine with this way too.
```
        const defaultTitle = e?.errorTitle ?? 'Something went wrong';
```

stevenhao:
thanks for contributing this change! do you have an .ipuz file handy for me to test with?

scouttyg:
@stevenhao I've attached a few iPuz and Puz files for comparison if they help, along with their sources:

[puz_and_ipuz_examples.zip](https://github.com/downforacross/downforacross.com/files/11222818/puz_and_ipuz_examples.zip)

## Sources:
1) https://crosshare.org/crosswords/GUNKcsoeee0gXGJNcICc/1847-british-literature-and-other-things
2) https://crosshare.org/crosswords/AsrgfbaS9P32f2x2w0eT/mama-mia
3) https://crosshare.org/crosswords/t4oUQn6DtLP9N62JtCdK/8-x-8
4) https://crosshare.org/crosswords/0kolHut2MOQWKxavia33/its-showtime

stevenhao:
cool, solved one of the ipuz's and it looked fine!

image

stevenhao:
thanks @scouttyg  so much for helping with this. closes #246.

PR #268, opened 2023-04-11
Store search query and filter options state in URL params

shintyl:
The changes in this PR persist the current search query and filter options of the Welcome page in the URL params, which allows users to link to a version of the Welcome page with a specific pre-populated search query and/or filter options. This is mainly designed to accommodate users that bookmark/pin a tab for [downforacross.com ](url) and are only interested in viewing New puzzles that fit a certain search string, and don't want to look through the full list of puzzles or enter a search string when they open/refresh the page.

# Demo
https://user-images.githubusercontent.com/14168765/231033457-31af2697-69ea-4021-9387-f011895c302e.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "react-helmet": "^5.2.0",
     "react-icons": "^3.10.0",
     "react-linkify": "^1.0.0-alpha",
-    "react-router-dom": "^4.1.2",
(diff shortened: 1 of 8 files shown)


stevenhao:
nice feature idea, and a pretty clean code diff too. thanks for contributing!

PR #270, opened 2023-05-02
Show dark mode toggle when dark mode is enabled

matsixfive:
The dark mode toggle is now persistent when dark mode is enabled because it is a beta feature. It should be hard to turn on but easy to turn off.


stevenhao:
good idea.

PR #273, opened 2023-05-25
Show a confirmation dialog before resetting the entire puzzle

craigkovatch:
I have, on several painful occasions, accidentally misclicked "Reset > Puzzle" rather than "Reset > Word". This PR adds a confirmation dialog before taking the Reset action at the Puzzle level 😅


craigkovatch:
@stevenhao just want to make sure you saw this :)

stevenhao:
Hey thanks for the reminder! I like this idea. Personally, i haven't reset a puzzle on accident before, but I'm sure it happens every once in a while so designing for that case is worthwhile.

Code change looks good, thanks for helping with this @craigkovatch!

Issue #247, opened 2022-08-15
Feature suggestion: Stats (best/average times)


czhang51:
It might be fun to show each player their best and average crossword completion times. (Without checks/reveal) Would love to help work on this if we think it is a useful feature.

stevenhao:
This sounds like a great suggestion! Feel free to open a PR :). A frontend-only PR with mock data is fine too if setting up the backend dev environment is troublesome.

PR #275, opened 2023-06-21
Add stats

dlakata:
Similar to https://github.com/downforacross/downforacross.com/issues/247, I thought it could be helpful to see the times of one's previous solves. As far as I could tell, the stats modules weren't used, so I repurposed them for this.

Let me know what you think!

crossword_stats

### DIFF
--- a/src/pages/Stats.tsx
+++ b/src/pages/Stats.tsx
@@ -66,7 +66,7 @@ const Stats: React.FC<{}> = () => {
                   <td>{avgSolveTime && formatMilliseconds(avgSolveTime)}</td>
                   <td>
                     {bestSolveTime && (
-                      <a href={`/beta/game/${bestSolveTimeGameId}`}>{formatMilliseconds(bestSolveTime)}</a>
+                      <a href={`/replay/${bestSolveTimeGameId}`}>{formatMilliseconds(bestSolveTime)}</a>
                     )}
                   </td>
                   <td>{avgCheckedSquareCount}</td>
@@ -93,7 +93,7 @@ const Stats: React.FC<{}> = () => {
               ({gameId, title, size, dateSolved, solveTime, checkedSquareCount, revealedSquareCount}) => (
                 <tr key={gameId}>
                   <td>
-                    <a href={`/beta/game/${gameId}`}>{title}</a>
+                    <a href={`/replay/${gameId}`}>{title}</a>
                     {` (${size})`}
                   </td>
                   <td>{dateSolved}</td>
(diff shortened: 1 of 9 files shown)


stevenhao:
Hey thanks for this PR! Love the idea, I'll take a look. The vercel preview is frontend-only, so it makes sense that your server-side changes aren't reflected.

stevenhao:
At first glance things look good -- maybe can add # players to History? + whether the game is in-progress or solved (and change the link to /replay/gid if it's completed)

dlakata:
I also was interested in getting the player count! But I was finding that the backend isn't aware of multiple users, where game_events don't actually have a user populated on them. Could you give me some pointers here?

If you'd prefer this, I can definitely rework things to handle incomplete games. My original thinking was that "stats" would only be useful for complete games, and https://downforacross.com/?complete=0&in_progress=1&new=0 could be used to see incomplete games. What do you think?

I updated all the links to the replay link!

src/pages/Stats.tsx
```diff
@@ -1,110 +1,112 @@
...
   useEffect(() => {
-    async function refresh() {
-      const allStats = await fetchStats();
-      setAllStats(allStats);
-    }
-    refresh();
-  }, []);
+    user.listUserHistory().then((history: any) => {
+      console.log(`keys: ${_.keys(history)}`);
+      fetchStats({gids: _.keys(history)}).then((s) => {
```
stevenhao, line 27: can we cap this at like, last 500 games? i'm not an expert in database performance but i'm worried that the backend could get slow if we try to do too many at once since the query involves joining with the game_events table (and there's no index on game event type afaik)

stevenhao:
i checked this out locally and it looks great! i left some small nitpicky comments inline, happy to merge after those are addressed.

as a follow-up after this is merged, i think it'd be a good idea to do a backfill and eagerly store & maintain the stats in the db to avoid having to do expensive queries. this would also allow us to track stats like number of distinct users + anything else we want to include in the stats page (there may be user feedback that we can act on).

dlakata:
I think I made all your suggested changes! and added another log line to help measure the performance of the endpoint. another index that might be good would be on puzzle_solves prefixed with gid rather than pid

idk what your database maintenance looks like, but yeah would be happy to help. when marking a puzzle solve, we could probably record the revealed/checked count. I wonder what stats other users would want (filtering different conditions, per time period, longest/current streak, day-of-week stats), and user feedback + performance could tell us what to improve

stevenhao:
great, thanks! merging this, i'll try to keep an eye on performance but not overly worried for now since there's limits on how many games are considered at once for stats

Issue #139, opened 2021-01-02
Don't allow uploading a previously-uploaded puzzle?


dpsanders:
Maybe it would be a good idea to not allow a puzzle to be uploaded if it has previously already been uploaded?

ewmson:
Now that the puzzles have been moved into postgres we *should* be able to enforce this with a constraint.

I think we want to allow non public (people testing or not wanting to publish) to have duplicates and only apply the constraint on the public one.

If we do go this route, there is a question of the past data as solves will be associated with different puzzle ids, but refer to the exact same puzzle. I suspect we can just start enforcing the constraint going forward to not worry about the past.

fg123:
In a similar realm, including other types of meta-data would be useful too, such as an "uploaded date" versus the puzzle date (i.e. for dated puzzles like NYT or LA Times). This would allow for retroactive uploading of puzzles but still allow front-end to sort by puzzle date, for example.

stevenhao:
duplicate of https://github.com/downforacross/downforacross.com/issues/15 (no pun intended)

Issue #15, opened 2018-07-11
Feature suggestion: duplicate detection


jakob223:
If someone uploads a puzzle that is already in the database, don't create multiple copies of it clogging up the home page.

stevenhao:
this should be doable by maintaining something like `puzzlehashes/:hash = {pid}` in the db.
- the hash should **not** depend on metadata, because different sources (e.g. cruciverb archive page, nyt archive-view, downloadacross, nyt puzzle-view all provide .puz files, but all name the puzzles differently)

stevenhao:
after detecting the duplicate, maybe we should a message like "this puzzle exists already" and show a link to /beta/play/:existingpid

PR #276, opened 2023-07-02
Prevent duplicate puzzles

dlakata:
For #139 and #15 

You'll need to create the index for this to take effect. I chose `30000` as the minimum `pid` to avoid enforcing unique constraints on existing data, and it appears to be an ID that'll be reached in the next week or two.

duplicate_screenshot

I created another table locally called `puzzles_test` and verified that the index seems to be used:
```
# select count(*) from puzzles_test;
 count  
--------
 424649
(1 row)

# \d puzzles_test
                         Table "public.puzzles_test"
    Column    |            Type             | Collation | Nullable | Default 
--------------+-----------------------------+-----------+----------+---------
 uid          | text                        |           |          | 
 pid          | text                        |           | not null | 
 pid_numeric  | numeric                     |           |          | 
 is_public    | boolean                     |           |          | 
 uploaded_at  | timestamp without time zone |           |          | 
 times_solved | numeric                     |           |          | 0
 content      | jsonb                       |           |          | 
Indexes:
    "puzzles_test_pkey" PRIMARY KEY, btree (pid)
    "puzzle_test_pid_numeric_desc" btree (pid_numeric DESC NULLS LAST)
    "unique_puzzle_test_content" UNIQUE, btree (md5(content ->> 'grid'::text), md5((content -> 'clues'::text) ->> 'down'::text), md5((content -> 'clues'::text) ->> 'across'::text)) WHERE pid > '30000'::text AND is_public
Check constraints:
    "puzzles_test_times_solved_check" CHECK (times_solved >= 0::numeric)

# explain analyze SELECT
pid
FROM puzzles_test
WHERE pid > '5'
AND is_public
AND md5(content->>'grid'::text) = md5(('[["A","B"],["C","D"]]'::jsonb)::text)
AND md5(content->'clues'->>'down'::text) = md5(('{"down": ["z","y","x","w","v"]}'::jsonb)->>'down')
AND md5(content->'clues'->>'across'::text) = md5(('{"across": ["a","b","c","d","3"]}'::jsonb)->>'across');
-------------------
Index Scan using unique_puzzle_test_content on puzzles_test  (cost=0.42..8.44 rows=1 width=6) (actual time=0.021..0.022 rows=0 loops=1)
   Index Cond: ((md5((content ->> 'grid'::text)) = 'c56da16f5cfea23f45f4dad64b283dc4'::text) AND (md5(((content -> 'clues'::text) ->> 'down'::text)) = '55f362a61d73771fc88b5987a990c553'::text) AND (md5(((content -> 'clues'::text) ->> 'across'::text)) = '1b4d1966f41032490216c0f73ec9c4bf'::text))
   Filter: (pid > '5'::text)
 Planning Time: 0.280 ms
 Execution Time: 0.049 ms
(5 rows)
```

### DIFF
--- a/server/api/puzzle.ts
+++ b/server/api/puzzle.ts
@@ -1,16 +1,29 @@
 import {AddPuzzleResponse, AddPuzzleRequest} from '@shared/types';
 import express from 'express';
 
-import {addPuzzle} from '../model/puzzle';
+import {addPuzzle, getMatchingPuzzle} from '../model/puzzle';
 
 const router = express.Router();
 
 router.post<{}, AddPuzzleResponse, AddPuzzleRequest>('/', async (req, res) => {
   console.log('got req', req.headers, req.body);
-  const pid = await addPuzzle(req.body.puzzle, req.body.isPublic, req.body.pid);
-  res.json({
-    pid,
-  });
+
+  try {
+    const pid = await addPuzzle(req.body.puzzle, req.body.isPublic, req.body.pid);
+    res.json({
+      pid,
+    });
+  } catch (err) {
+    if (err.code == '23505') {
+      const pid = await getMatchingPuzzle(req.body.puzzle);
+      console.log(`Unique constraint violated, attempt pid (${req.body.pid}) for existing pid (${pid})`);
+      res.status(400).json({duplicatePuzzle: pid});
+    } else {
+      console.error(err.message);
+      console.error(err.code);
+      res.status(400).json();
+    }
+  }
 });
 
 export default router;
--- a/server/model/puzzle.ts
+++ b/server/model/puzzle.ts
@@ -134,6 +134,29 @@ export async function addPuzzle(puzzle: PuzzleJson, isPublic = false, pid?: stri
   return pid;
 }
 
+export async function getMatchingPuzzle(puzzle: PuzzleJson): Promise<string | undefined> {
+  const {rows} = await pool.query(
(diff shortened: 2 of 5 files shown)


dlakata:
> Could we make the constraint only apply to public puzzles? Unlisted puzzles should allow duplicates

I think the `WHERE pid > '30000' AND is_public` clause should handle this: https://www.postgresql.org/docs/current/indexes-partial.html

> do we need to materialize the md5 hashes in db (and add an index on it) as an optimization? I'm not too familiar w/ using md5 but this looks like an unindexed query that could get pretty slow...

I think `CREATE UNIQUE INDEX unique_puzzle_content` should add the index, which the `Index Scan using unique_puzzle_test_content` test in the description used. Based on https://www.postgresql.org/docs/7.3/indexes-functional.html, I think we don't need to materialize, and postgres will store the index results. But I'm new to functional indexes, so you should verify my understanding!

stevenhao:
ok so i think it's probably the case that people aren't intentionally creating duplicates, so a database unique index doesn't feel necessary. i like the frontend changes though,               `Play the puzzle <a href={url}>here</a>` makes a lot of sense.

stevenhao:
since this PR is a year old i'm gonna go ahead and close it but i'd be happy to merge just the "detect duplicate & link the original" frontend logic but still giving the option of uploading it anyway (e.g. maybe there's a reason to do that -- perhaps the new puzzle title/metadata is a significant change & worth reuploading for)

PR #284, opened 2023-08-10
Implement link previews

shintyl:
This PR implements a feature that generates link previews (also known as "embeds") for DFAC game URLs (e.g. `https://downforacross.com/beta/game/1234567-joct`) sent in platforms such as Discord and Slack. These previews are generated by "link unfurling" crawlers on these platforms that visit the URLs posted in sent messages and search for certain tags.

Using [Vercel Middleware](https://vercel.com/docs/concepts/functions/edge-middleware), requests can be intercepted before the response is served, and if the request URL is a game URL, and if user-agent string is suggestive of a crawler, the crawler can be redirected while human users are unaffected. Crawlers are redirected to an [API endpoint ](https://github.com/shintyl/downforacross.com/blob/f11bf29120d8cdaff9575aef901995e9f5a6e7da/server/api/link_preview.ts) which returns a barebones HTML document containing [Open Graph Protocol](https://ogp.me/) tags (for that specific game), whose content populates the link preview.

Here's what the link previews look like in Discord and Slack (ignore the URLs):

image
image

Simpler approaches, like mutating the DOM in the client-side code to include the tags, unfortunately don't work because none of these crawlers actually seem to execute JavaScript, and only see [index.html](https://github.com/downforacross/downforacross.com/blob/f11bf29120d8cdaff9575aef901995e9f5a6e7da/public/index.html)'s initial state. Another approach like moving to server-side rendering would probably take a lot of work.

Note that link previews for the Vercel preview deployment won't be generated because I don't think API changes are reflected until merge.

### DIFF
--- a/utils/link_preview_util.ts
+++ b/utils/link_preview_util.ts
@@ -1,0 +1,8 @@
+const linkExpanderUserAgents = [
+    'Discordbot',
+    'Slackbot-LinkExpanding'
+]
+
+export function islinkExpanderBot(userAgent: string) {
+    return linkExpanderUserAgents.some(ua => userAgent.includes(ua))
+}
--- a/utils/link_preview_util.ts
+++ b/utils/link_preview_util.ts
@@ -1,8 +1,13 @@
-const linkExpanderUserAgents = [
-    'Discordbot',
-    'Slackbot-LinkExpanding'
-]
+const linkExpanderUserAgentSubstrings = {
+    Discord: 'Discordbot',
+    Slack: 'Slackbot-LinkExpanding',
+    FB_Messenger: 'facebookexternalhit',
+};
+
+export function isFBMessengerCrawler(userAgent: string) {
+    return userAgent.includes(linkExpanderUserAgentSubstrings.FB_Messenger)
+}
 
 export function islinkExpanderBot(userAgent: string) {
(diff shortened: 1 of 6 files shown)


utils/link_preview_util.ts
```diff
@@ -0,0 +1,8 @@
+const linkExpanderUserAgents = [
+    'Discordbot',
+    'Slackbot-LinkExpanding'
```
stevenhao, line 3: are there other strings we need to include here? i remember looking into this in the past to support fb messenger link previews, though i'm guessing it's a different protocol? https://ogp.me/

stevenhao:
This looks really good! Ready to merge whenever; if it's not much work to support link previews on fb messenger then we can try to get that included first.

utils/link_preview_util.ts
stevenhao: i found this thingy which might be helpful https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json

shintyl: So interestingly, it seems like Messenger only supports title + thumbnail in its link previews:
image

So as a hack I ended up just cramming everything into the title (line breaks don't seems to work)
image

stevenhao: nice, looks great!!

stevenhao:
shipped!! 
image

PR #285, opened 2023-08-10
Fix Messenger link preview formatting

shintyl:
Fix minor issue seen in https://github.com/downforacross/downforacross.com/pull/284#issuecomment-1673702485 where Messenger link previews contained an unnecessary vertical bar if the description was empty.

Missing descriptions should look like this now:
image


shintyl:
Did this change get deployed? I'm still seeing the extra vertical bar

PR #287, opened 2023-08-27
Show number of squares filled by each user in Color Attribution Mode

mitchellvitez:
Inspired by https://github.com/downforacross/downforacross.com/issues/286

Shows the number of squares filled in by each player at each point on the Replay timeline

When scrubbed to the end of the Replay timeline, this will be the number of squares solved by each player

downforacross squarecounts

downforacross squarecounts

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,10 @@ If you notice a bug or have a feature request, feel free to open an issue.
    `git clone https://github.com/downforacross/downforacross.com.git`
    `cd downforacross.com`
 
-3. Use node v14
+3. Use node v18
    `nvm install`
    `nvm use`
-   `nvm alias default 14` (optional)
+   `nvm alias default 18` (optional)
 
 4. Install Dependencies
    `yarn`
--- a/src/components/Player/Player.js
+++ b/src/components/Player/Player.js
@@ -308,6 +308,39 @@ export default class Player extends Component {
     );
   };
 
+  renderColorAttributionCounts() {
+    if (!this.props.colorAttributionMode || this.state.error) {
+      return null;
+    }
+
+    // map from displayName to number of squares solved by that user
+    const counts = {};
+    this.props.grid.forEach((row) => {
+      row.forEach((cell) => {
+        if (cell.user_id) {
+          counts[cell.user_id] = (counts[cell.user_id] || 0) + 1;
+        }
+      });
+    });
+
+    if (Object.keys(counts).length === 0) {
+      return null;
+    }
+
+    return (
+      <div style={{marginTop: 24}}>
+        <strong>Squares filled by user</strong>
+        {Object.entries(counts)
+          .sort((x) => x[1])
+          .map(([userId, count]) => (
+            <div style={{color: this.props.users[userId]?.color}}>
+              {this.props.users[userId]?.displayName} - {count}
+            </div>
+          ))}
+      </div>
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -17,10 +17,10 @@ If you notice a bug or have a feature request, feel free to open an issue.
    `git clone https://github.com/downforacross/downforacross.com.git`
    `cd downforacross.com`
 
-3. Use node v14
+3. Use node v18
    `nvm install`
    `nvm use`
-   `nvm alias default 14` (optional)
+   `nvm alias default 18` (optional)
```
mitchellvitez, line 23: drive-by change since the `.nvmrc` had `18` in it

stevenhao:
Thanks for this PR! I'm not sure whether this feature would be redundant with the color attribution mode. And I'm also not sure if it needs to be always-showing, given that only a subset of users would find it helpful.

Perhaps we could make this show up only if color attribution mode is enabled? And render the usernames with colors as well?

mitchellvitez:
Agreed that only a subset of users would find this helpful/interesting

For us, we often have 5–10 people playing so the colors can be a little hard to distinguish. I'd also argue it's not totally redundant because it saves us from having to count up squares manually to find this user ordering.

I think showing up only in color attribution mode (with user colors) is a great idea, just made that change:

Screenshot 2023-09-03 at 10 17 41 AM

ekilah:
Excited for this!

jsphwllng:
This is exactly what I hoped for with my issue. Thanks for this :sparkle:

src/components/Player/Player.js
```diff
@@ -308,6 +308,39 @@ export default class Player extends Component {
...
+
+    if (Object.keys(counts).length === 0) {
+      return null;
+    }
+
+    return (
+      <div style={{marginTop: 24}}>
+        <strong>Squares filled by user</strong>
+        {Object.entries(counts)
+          .sort((x) => x[1])
```
stevenhao, line 334: not sure if this sort call is correct, javascript is a bit weird

stevenhao:
Sorry for being slow to review this! I think if we fix the sort thingy we can get it merged. Thanks for your patience :)

jsphwllng (approved):

mitchellvitez:
No worries about timing


I took a stab at making the sorting more clear. If that's not what you meant, can you say more about what you want fixed with the sort?

I checked that

```js
typeof(Object.entries(counts)[0][1])
```

was `number` in the original

stevenhao:
> No worries about timing
> 
> I took a stab at making the sorting more clear. If that's not what you meant, can you say more about what you want fixed with the sort?
> 
> I checked that
> 
> ```js
> typeof(Object.entries(counts)[0][1])
> ```
> 
> was `number` in the original

I think the way you're invoking the function might not be right. The docs here should be helpful https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

mitchellvitez:
Ah you're right, great catch! I assumed `sort` took a property to sort on, not a comparison function

Should be fixed now

mitchellvitez:
Ah you're right, great catch! I had assumed `sort` would take a property to sort on, not a comparison function

Should be fixed now

stevenhao:
awesome work!!

Issue #288, opened 2023-09-01
Nav icons are too small to click on small screens


Nyiriland:
On my iPhone 13 mini (375px width), the last 2 nav buttons are impossible to click as they shrink to fit the space, I'm assuming due to them being svgs contained in a flex box. See screenshot below.

**Suggested solution**
Enable horizontal scrolling on the header for overflow, and don't allow icons to shrink.

<img width="883" alt="image" src="https://github.com/downforacross/downforacross.com/assets/3835008/54630ecc-9e40-48b7-b7d3-cc3ee77c1d25">

PR #289, opened 2023-09-02
add menu toggle for mobile

christianscarlett:
### Summary
In response to #288 I tried my hand at a solution by adding a menu expansion toggle to reveal the icons that were being shrunk.

### Changes Made
- Added state variable to toggle between menus
- Added icon that shows/hides the alternative menu items on tap
- Called to update clock when mounted (otherwise it would show 00:00 for a second when toggling between menus)

### Test Plan
Show toggle hides and reveals the correct menu items as expected, and that the menu items still work:
ezgif com-video-to-gif

### DIFF
--- a/src/components/Game/Game.js
+++ b/src/components/Game/Game.js
@@ -23,6 +23,7 @@ export default class Game extends Component {
       vimMode: false,
       vimInsert: false,
       colorAttributionMode: false,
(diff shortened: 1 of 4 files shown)


stevenhao:
Sorry I missed this earlier. This looks great. Thanks for contributing!

PR #292, opened 2023-12-02
Local storage for vim mode

### LINKED ISSUE #291: Save and load previously toggled vim mode settings
This is a minor enhancement to save vim mode settings across sessions so that the user doesn't have to enable it every time.

One easy fix would be a minor change to `src/components/Game/Game.js` to fetch and update a vimMode field from `localStorage`.

vasumv:
Resolves #291.

This PR adds local storage for the vim mode toggle to be used in the state of the `Game` component. This avoids the need to enable Vim mode every time the crossword is loaded.

Tested locally and on preview.

### DIFF
--- a/src/components/Game/Game.js
+++ b/src/components/Game/Game.js
@@ -11,6 +11,8 @@ import Toolbar from '../Toolbar';
 import {toArr} from '../../lib/jsUtils';
 import {toHex, darken, GREENISH} from '../../lib/colors';
 
+const vimModeKey = 'vim-mode';
+
 // component for gameplay -- incl. grid/clues & toolbar
 export default class Game extends Component {
   constructor() {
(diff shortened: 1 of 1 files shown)


src/components/Game/Game.js
```diff
@@ -29,9 +31,11 @@ export default class Game extends Component {
 
   componentDidMount() {
     const screenWidth = window.innerWidth - 1; // this is important for mobile to fit on screen
+    const vimMode = JSON.parse(localStorage.getItem(vimModeKey)) || false;
```
stevenhao, line 34: this might need a try catch just to be safe -- otherwise a bad value in localStorage could brick the site

stevenhao (commented):
great idea! left a comment, good to merge after that is addressed

src/components/Game/Game.js
vasumv: good point! added the try/catch

stevenhao:
great! thanks for the PR!

PR #293, opened 2023-12-04
Select clues in Vim Mode with clue number

### LINKED ISSUE #294: Add jumping to specific clues in Vim Mode
Similar to jumping to specific line numbers in Vim, it would be nice to select a specific clue using Vim syntax (e.g. ":55" or ":55a" for 55 across). 

This could be done using an additional mode (similar to `vimInsert`) called `vimCommand` that activates when the ":" key is typed. Then, a buffer of keys can be populated until the user hits Enter. 

A box in the bottom left/right displaying the buffer of keys could also be shown when command mode is activated.

PR: #293

vasumv:
This PR adds functionality to select clues in Vim Mode using line number selection syntax (e.g. ":56" or ":56a"). Defaults direction to across.

Adds a new Vim Command mode activated by ":" that populates a command bar component. When pressing enter, the keys are parsed for a clue number and direction. Current matching is limited to just selecting clues.

Visualization of keys pressed in command mode in the bottom in a blue box. 
xword

Resolves #294


stevenhao:
sorry i think i forgot to review this. let me take a look!

src/components/Game/Game.js
```diff
@@ -173,9 +176,21 @@ export default class Game extends Component {
...
   };
 
+  handleVimCommand = () => {
+    this.setState({
+      vimCommand: true,
+    });
+  };
+
+  handleVimCommandBuffer = (key) => {
+    this.setState({vimCommandBuffer: [...this.state.vimCommandBuffer, key]});
```
stevenhao, line 186: can we should check if key is special key? e.g. backspace, shift & other modifiers, esc, who knows what. seems like esc works already but backspace & shift don't.

stevenhao:
just tested this on the vercel preview deployment. very neat!

left some inline comments about special keys like backspace & delete not working in the command buffer. good to merge after that's addressed!

vasumv:
Thanks @stevenhao! Yeah, should definitely handle special keys. I'm currently out of town but will make an update in a week or so.

vasumv:
@thkim1011 and I have refactored the PR a bit to add a new component for the Vim Command Bar. This should handle special keys as well. @stevenhao please take a look when you have the chance, thanks!

stevenhao:
Hey @vasumv sorry for the delay. i tested it and noticed that insert mode vs normal mode look the same. can we make insert mode show up in status bar? this may be due to a toolbar redesigns, not sure.

image

vasumv:
Thanks for taking a look @stevenhao ! Here is an updated version with the insert mode displayed on the status bar (bottom left). 
insertmode

stevenhao:
looks good!

PR #295, opened 2023-12-07
Added function in ClueText to determine whether text should be italic…

zhiyi-zhang-duke:
This code change adds a `shouldItalicize` function as well as changes `ClueText.js` into a react component. If regex matches twice, we know that the text needs to be italicized.  Here's a screenshot of the new component in action:
image

### DIFF
--- a/src/components/Player/ClueText.js
+++ b/src/components/Player/ClueText.js
@@ -1,44 +1,73 @@
-import React from 'react';
+import React, {Component} from 'react';
 
-function decodeHtml(htmlText) {
-  let text = document.createElement("textarea");
-  text.innerHTML = htmlText;
-  return text.value;
-}
-
-export default ({text = ''}) => {
-  const parts = [];
-  while (text.length > 0) {
-    const s = text.indexOf('<i>');
-    const e = text.indexOf('</i>');
-    if (s === 0 && e !== -1) {
-      parts.push({
-        text: text.substring(3, e),
-        ital: true,
-      });
-      text = text.substring(e + 4);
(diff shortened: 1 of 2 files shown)


stevenhao (commented):
Thanks for the PR!

Two comments:
1. can we keep this as function component instead of a class component?
2. i think there are a lot of false positives for clues that start & end with quotation marks, so as written i don't think the behavior is strictly correct. Perhaps we can change the rule to checking if the clue text is encased by 2 quotes (`""x""` instead of `"x"`)?
3. since there is already some logic involving parsing out `<i>` tags for italics, maybe we should avoid mixing the control flow with that? e.g. ```if entire clue is italic: ... else parse out <i> parts```

zhiyi-zhang-duke:
> 1. can we keep this as function component instead of a class component?
Wow I totally missed it was already a functional component. Sorry about that, it's restored!
> 2. i think there are a lot of false positives for clues that start & end with quotation marks, so as written i don't think the behavior is strictly correct. Perhaps we can change the rule to checking if the clue text is encased by 2 quotes (`""x""` instead of `"x"`)?
Great point, I made the regex do exactly what you said and only check for two double quotes at the beginning and end of the string.
> 3. since there is already some logic involving parsing out `<i>` tags for italics, maybe we should avoid mixing the control flow with that? e.g. `if entire clue is italic: ... else parse out <i> parts`
Ah I actually glossed over that logic instead of really trying to understand it. 😅 I incorporated the control flow you suggested.

stevenhao:
great!

PR #299, opened 2023-12-16
Persist pencil color

OhPointFive:
This PR automatically saves the user's pencil color, so it will no longer reset to gray on page reload or when starting a new puzzle.


stevenhao:
looks good!

PR #300, opened 2023-12-16
Added a default username that persists across games when set

zhiyi-zhang-duke:
This code change adds a `default_username` to the localStorage that is only set when the user updates their username to something that is not a randomly generated name. This value is then checked when other games are loaded and if it is set, is used instead of the randomly generated name.

Here's the state of local storage after a new game is created:
image

Here's the state of local storage after the username is updated:
image

Here's the state of a new game after the `username_default` is set:
image


stevenhao:
looks good!

Issue #301, opened 2024-01-04
Underlined clues don't render correctly


zhiyi-zhang-duke:
If you look at one of the recent puzzles:
NY Times, Thursday, January 4, 2024
51 across is rendered as:
<img width="232" alt="image" src="https://github.com/downforacross/downforacross.com/assets/68999131/84b2bf78-ec82-424f-ad5e-6ba1c539d738">
64 across:
<img width="217" alt="image" src="https://github.com/downforacross/downforacross.com/assets/68999131/d92d6fb1-055c-4033-8da3-4517378faba7">
I'm guessing they're just supposed to just be underlined.

lllloyd:
Here's a console command to fix it for today!

```
let els = document.getElementsByClassName("clues--list--scroll--clue--text");
for (const el of els) {
let e = document.createElement("div");
e.innerHTML = el.firstChild.innerText;
el.removeChild(el.firstChild);
el.appendChild(e);
}
```

zhiyi-zhang-duke:
I believe this can be closed!

PR #302, opened 2024-01-07
feat: improve markup support

brownie-in-motion:
re-implementation of components/Player/ClueText.js that

1. parses HTML in clues in a more robust way and
2. allows additional markup to be easily supported.

additionally, this automatically handles HTML entities, removing the need for textarea hacks

example puzzle for testing: [markup-test.zip](https://github.com/downforacross/downforacross.com/files/13854967/markup-test.zip)


src/components/Player/ClueText.ts
```diff
@@ -0,0 +1,75 @@
...
+  // case where we should italicize the whole clue
+  if (text.startsWith('""') && text.endsWith('""')) {
+    return createElement('i', {}, text.slice(1, -1))
+  }
+
+  // fast path for text with no HTML and no entities
+  if (!text.match(/[<>]|&[^;]+;/)) return createElement('span', {}, text)
+
+  // otherwise, parse HTML and render allowed elements
+  const allowed = ['em', 'strong', 'u', 'i', 'b', 'sup', 'sub']
```
stevenhao, line 72: have to admit, i'm a bit scared of this haha. what if a script tag somehow makes it through?

stevenhao: is there a way weird props / malformed xml tags could do damage, even if it passes the tag whitelist?

brownie-in-motion: I think it's fine because we create react nodes based on the template's tree, and every call to `createElement` can only have a name in that list.
```ts
  if (tree.name !== undefined && allowed.includes(tree.name)) {
    return createElement(tree.name, {}, ...children)
  } else {
    return createElement(Fragment, {}, ...children)
  }
```
but let me know if there's something i'm overlooking

brownie-in-motion:
(fixes #301)

src/components/Player/ClueText.ts
stevenhao: i see, so props are all filtered out.

brownie-in-motion: instead of *filtering* the markup for dangerous content, we recreate the html. so this gives the application total control over what elements can be emitted (specifically, those in the allow list) and ensures that no props can be set.

zhiyi-zhang-duke:
Can we get some screenshots of the code in action for some of the allowed html elements? I'm curious what they look like. I was going to checkout your branch myself but figured it might be easier for everyone involved if we can see it from your local machine.

brownie-in-motion:
Sure, here is what the example puzzle I uploaded looks like with the changes (left) compared to the current version (right)
image

zhiyi-zhang-duke:
Thanks, the screenshots really help! @stevenhao would you rather try deploying this feature with a feature flag?
https://posthog.com/tutorials/react-feature-flags

stevenhao:
hey sorry for dropping this. reviewed again & i think it's probably not going to be a security risk! thanks for your patience :)

Issue #304, opened 2024-01-18
Feature request: Clickable clues in chat


fishy15:
It would be nice if the user types in something like "30 across" in chat, and someone else can click on it to go directly to the clue. Not sure what should be supported, if support for other languages or what phrases exactly, but an initial baseline implementation would be to match any substrings in the form of a number followed by across/down (case-insensitive).

fishy15:
Closed by #305

PR #305, opened 2024-01-23
Clickable clue references

fishy15:
Allows user to click on clue references and go to the matching clue. Implements #304.

Would definitely appreciate feedback on UI, I simply just converted it into a button to make it stand out, but something else could be done as well.

### DIFF
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -334,7 +334,8 @@ export default class Chat extends Component {
       return defaultPattern;
     }
 
-    const isAcross = clueref[2][0] === 'a';
+    const directionFirstChar = clueref[2][0];
+    const isAcross = directionFirstChar == 'a' || directionFirstChar == 'A';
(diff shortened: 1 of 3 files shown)


src/components/Chat/Chat.js
```diff
@@ -310,7 +311,7 @@ export default class Chat extends Component {
             {token.type === 'emoji' ? (
               <Emoji emoji={token.data} big={bigEmoji} />
             ) : token.type === 'clueref' ? (
-              token.data // for now, don't do anything special to cluerefs
+              this.renderClueRef(token.data)
```
stevenhao, line 314: interesting that this code already existed! i remember wanting this feature before but i guess never pushed it through

fishy15:
I believe the logic works right now, you can review that if you want. But if you have any design ideas for the component, you can implement those. Currently I am thinking of making it look like a hyperlink.

stevenhao:
image
styling seems alright, i think the next step would be to make the feature more discoverable. any ideas?

stevenhao:
merging this now

PR #308, opened 2024-01-31
[fix] Autocheck intermittently doesn't work 

justindhill:
## Summary
Fixes an issue where autocheck intermittently doesn't work on a cell due to out of order processing of `updateCell` and `check` events.

## Description
Currently, autocheck is implemented by the frontend adding two game events in close succession, an `updateCell` event and a `check` event whose scope is constrained to the cell coordinates of the updated cell. This makes sense conceptually, since socket.io guarantees ordering of messages regardless of the transport it's using. Problems can arise due to minute variations in how long it takes to process an event. Occasionally, the `check` event finishes processing before the `updateCell` event. Clients then receive the events out of order, causing the check to be performed on an unpopulated cell.

## Solution
This problem really only arises because of how close the events are fired to one another. Other events don't follow this pattern, so are far less likely to experience this problem. Additionally, for other events, ordering may not be as important. Therefore, instead of trying to solve the general case of processing finishing in-order, we can solve this particular problem by adding an `autocheck` parameter to the `updateCell` event and not firing a `check` event at all. Processing of `check` events remains the same, so explicit checks of cells, words, and puzzles are unaffected.

## Before / After
Untitled 2 Untitled


flackr:
The code looks good and this seems to work pretty well for me, any chance we can get this merged to get auto-check working flawlessly?

flackr:
@justindhill can you merge master into the branch so I can play test this with all of the other changes that have landed since?

stevenhao:
@justindhill thanks for submitting this fix! if you could address merge conflicts i'll review it

justindhill:
All set @stevenhao

flackr:
Anything I can do to help? @stevenhao

stevenhao:
ok finally reviewed! logic looks good. agree that solving event ordering can be left for the future.

PR #311, opened 2024-02-13
Use Sequences for GID, PID

christianscarlett:
Adds 2 sequences for generating GIDs and PIDs.
https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-sequences/

Migrations needed to be run before this is deployed are in server/sql/create_id_counters.sql

I tried to clean up all firebase references to GID and PID, but there was one in Replays I wasn't sure what to do with.

There are only a couple firebase references left in the client code:
1. Replays
2. Composition
3. User auth
4. Battle

I'm not sure any of these are used, maybe we can clean these up too.

Tested uploading a puzzle and creating + opening games on local frontend, backend, and db:

ScreenRecording2024-02-14at9 44 56PM-ezgif com-video-to-gif-converter


server/sql/create_id_counters.sql
```diff
@@ -0,0 +1,7 @@
+-- psql < create_id_counters.sql
+
+CREATE SEQUENCE IF NOT EXISTS gid_counter START 100000000;
+CREATE SEQUENCE IF NOT EXISTS pid_counter START 100000000;
```
christianscarlett, line 4: Just chose 10,000,000 because it seems higher than any of the existing GIDs

PR #319, opened 2024-04-25
Add share score when game is finished

catherynl:
when game is finished, this appears in chatbox:
image

what gets copied to clipboard:
```
I solved NY Times, Wednesday, April 24, 2024 in 00:28!

http://localhost:3020/beta/play/34343
```
note that this is a link to create a new game with the same puzzle, not a link to the existing game.

### DIFF
--- a/src/api/create_game.ts
+++ b/src/api/create_game.ts
@@ -1,7 +1,5 @@
 // ========== GET /api/puzzlelist ============
 
-// eslint-disable-next-line import/no-extraneous-dependencies
-import qs from 'qs';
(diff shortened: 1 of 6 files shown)


stevenhao:
probably should link to the game too? and maybe `We solved` if there were collaborators

catherynl:
> probably should link to the game too?
I decided against it because having two links in one message would be too confusing. there's already a way to link to the game itself -- wdyt?

> maybe `We solved` if there were collaborators
I like that! will update

PR #322, opened 2024-05-15
Create and focus an invisible input field on click.

### LINKED ISSUE #192: Can't show keyboard on iPad
Sorry, I'm new here, but I can't figure out how to show the keyboard when using the site on an iPad.  I can select squares, but cannot enter any text, because there is no keyboard.  Is there some hidden feature to show a keyboard / enter text?

flackr:
This fixes #192, allowing using the non-mobile interface on devices without keyboards as having a focused input field triggers the virtual keyboard on devices that need it.

### DIFF
--- a/src/components/Grid/Cell.tsx
+++ b/src/components/Grid/Cell.tsx
@@ -185,13 +185,11 @@ export default class Cell extends React.Component<Props> {
 
   handleClick: React.MouseEventHandler<HTMLDivElement> = (e) => {
     e.preventDefault?.();
-    e.stopPropagation?.();
     this.props.onClick(this.props.r, this.props.c);
   };
 
   handleRightClick: React.MouseEventHandler<HTMLDivElement> = (e) => {
     e.preventDefault?.();
-    e.stopPropagation?.();
     this.props.onContextMenu(this.props.r, this.props.c);
   };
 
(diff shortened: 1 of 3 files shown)


src/components/Grid/Cell.tsx
```diff
@@ -255,22 +253,6 @@ export default class Cell extends React.Component<Props> {
...
-          onTouchEnd={(e) => {
-            if (e.changedTouches.length !== 1 || e.touches.length !== 0) return;
-            const touch = e.changedTouches[0];
-            if (
-              !this.touchStart ||
-              (Math.abs(touch.pageX - this.touchStart.pageX) < 5 &&
-                Math.abs(touch.pageY - this.touchStart.pageY) < 5)
-            ) {
-              e.preventDefault();
-              onClick(this.props.r, this.props.c);
```
flackr, line 271: It looks like this was trying to detect clicks without the 300ms click delay, but this can be done with the touch-action: manipulation property which I've added to the css, which results in immediate click events after touchdown and touchup.

src/components/Grid/Cell.tsx
```diff
@@ -185,13 +185,11 @@ export default class Cell extends React.Component<Props> {
 
   handleClick: React.MouseEventHandler<HTMLDivElement> = (e) => {
     e.preventDefault?.();
-    e.stopPropagation?.();
```
flackr, line 188: The stopPropagation here is removed so that the click can also focus the input element in GridControls when the event propagates up to that element.

flackr:
I've ensured that the desktop version still plays fine from a computer as well as now supporting triggering touch device keyboard, playing it pretty much every day from the preview link. Can I do anything to help get this landed?

stevenhao:
hey sorry for the delay for review. i don't have an ipad so i haven't been able to work on this, appreciate your help! i tried out the preview link on desktop; when clicking on a cell there seems to be some subtle flickering that doesn't happen on main site, maybe there's some :focus thing going on?

flackr:
> hey sorry for the delay for review. i don't have an ipad so i haven't been able to work on this, appreciate your help!

You can also test it on an iPhone if you have one by clicking the font-like icon in the url bar and telling it to load the desktop site.

> i tried out the preview link on desktop; when clicking on a cell there seems to be some subtle flickering that doesn't happen on main site, maybe there's some :focus thing going on?

Thanks for catching. Yeah we were using :focus on the grid-control but focus is moved within. I've updated it to `:focus-within` which is well supported and fixes the issue.

flackr:
@stevenhao anything more I can do to help? Here's a screenshot of me using the desktop interface with keyboard working on my iphone:

typing

Note that I've forced it to load the desktop version to show that it works. It will still load the mobile interface by default on the phone.

stevenhao:
nice! looks great.

PR #323, opened 2024-05-19
Add autocheck button to mobile toolbar.

flackr:
This adds the autocheck button to the expanded menu on the mobile toolbar with updated styling to make its state visible above the blue menubar.

### DIFF
--- a/src/components/Toolbar/css/index.css
+++ b/src/components/Toolbar/css/index.css
@@ -99,6 +99,13 @@ code {
   color: var(--main-blue) !important;
 }
 
+.toolbar--mobile .toolbar--autocheck {
+  color: var(--main-gray-2);
+}
+.toolbar--mobile .toolbar--autocheck.on .fa-check-square {
+  color: white !important;
+}
+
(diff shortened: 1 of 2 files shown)


flackr:
My friends and I really like the autocheck feature. If you find the second page of the toolbar is getting too crowded we could always either add a menu or make the expand button move between two or more additional menu panes.

5hubh4m:
I love auto check, so I would really appreciate it if this PR was merged into the main deployment.

However, the current implementation seems to have a bug where there’s a small chance an entered letter does not auto check and remains black. If you backspace and type it again it auto checks correctly.

flackr:
> However, the current implementation seems to have a bug where there’s a small chance an entered letter does not auto check and remains black. If you backspace and type it again it auto checks correctly.

This is true, however this is an issue with autocheck already (outside of this PR) and should be treated as a separate issue / pull request to fix.

5hubh4m:
Yes I realised that after I looked at the commit!

stevenhao:
lgtm!

PR #324, opened 2024-05-20
fix parrot emojis and remove dead urls

### LINKED ISSUE #321: emoji urls dead
seems like the emojis were added a while back, and many of the image urls are no longer working. is there any reason to implement the emojis like this rather than adding the files directly as resources? i'm happy to clean it up or just comb through and remove dead links but figured i should ask first

abbymartin:
fixes #321


stevenhao:
thanks for fixing!

PR #327, opened 2024-06-02
Fix backspace behavior on mobile.

### LINKED ISSUE #317: Backspace doesn't consistently go to previous cell on Android
Hello, love the website! I noticed an annoyance on Android that doesn't seem to exist on desktop.

Expectation: Backspace will delete a letter and highlight the previous cell, so that I can spam backspace to delete entire words.

Behavior: Backspace deletes the letter, but the same cell (now empty) remains highlighted. Further taps of backspace accomplish nothing.

Video of the bug: https://photos.app.goo.gl/jF3jJNjKbCKPwTxFA 

Android 14; Pixel 6 Build/AP1A.240405.002
Chrome 124.0.6367.54

I reckon this code is relevant: https://github.com/downforacross/downforacross.com/blob/98004d8f78a77ddbbdd8f50526c65e51a995756f/src/components/Player/GridControls.js#L368-L375

tarehart:
From reading the code, I've inferred that the input box has a well-defined initial state that we always reset to: It has a value of "$", and the cursor is always at the end.

In cases where the cursor is NOT at the end as expected, we see two error cases:
- User enters a letter, the change handler fires with input like "a$", and "$" is incorrectly entered in the grid.
- User attempts to backspace, and the change handler does not fire at all because the cursor was at position 0 (Fixes #317)

I've developed and tested this fix using an android emulator attached to a chrome debug window.

### DIFF
--- a/src/components/Player/MobileGridControls.js
+++ b/src/components/Player/MobileGridControls.js
@@ -432,12 +432,26 @@ export default class MobileGridControls extends GridControls {
     }
   };
 
+  /**
+   * There are hidden input boxes on the page, this handler listens for changes and then relays the inferred
+   * user input to the crossword grid. The input box has a well-defined initial state that we always reset to:
+   * It has a value of "$", and the cursor is always at the end.
+   *
+   * By comparing with this initial state, we can infer what the user did, i.e. if the new value is "$a" they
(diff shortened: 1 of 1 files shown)


stevenhao:
thanks for tracking this down!

tarehart:
Thanks for reviewing!

Issue #313, opened 2024-02-17
Add Shortcut to Copy on Grid


kumar-ish:
Sometimes I want to look up a word after revealing it and would like the option of being able to copy it so I can paste it and find out its meaning on {search_engine}.

PR #337, opened 2024-11-09
Shortcut to Copy on Grid 

chbrky:
Allows user to right-click on revealed clue to automatically paste answer and clue into search engine. Implements Add Shortcut to Copy on Grid #313

### DIFF
--- a/src/components/Grid/Grid.tsx
+++ b/src/components/Grid/Grid.tsx
@@ -36,6 +36,9 @@ export interface GridProps {
   editMode: boolean;
   frozen: boolean;
 
+  // Clue lookup
+  clues: {across: string[]; down: string[]}; //figure out type
+
   // callbacks
   onChangeDirection(): void;
   onSetSelected(cellCoords: CellCoords): void;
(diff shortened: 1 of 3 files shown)


stevenhao:
thanks for the PR!
1. i think right-click to copy is fine (maybe show a little toast saying clue & text "Copied to clipboard") but opening google search in a new tab is a bit much... 
2. maybe the trigger should be right clicking the clue in the clue list, instead of on the grid? the grid already uses right click for pings.

chbrky:
Thanks for reviewing the PR!
1. What if we still do a Google search but don't focus on the new tab created? I think it would be easier for the user if they had a bunch of answers/clues they wanted to look up rather than having to repeatedly switch tabs.
2. I totally agree with the right-click already being used with the ping on the grid, and I don't like the multi-use, but I believe copying from the clue list isn't exactly intuitive (I tried to implement it while working on it early). The ping still works, except when a clue is revealed, then it will search for the answer. My question is how often are people pinging revealed answers?

Issue #212, opened 2021-10-11
upgrade chat ui


stevenhao:
this lib looks p good https://brandonmowat.github.io/react-chat-ui/demo/

PR #339, opened 2024-11-12
Chat UI Upgrades

chbrky:
Chat UI Upgrades. Added message grouping, additional styles for chat messages, and fixed chat spacing issues. Implements #212


chbrky:
@stevenhao

PR #344, opened 2025-03-24
Allow username editing on mobile

cameronblandford:
I do the crossword every morning with my friends. When I do it on mobile, I can't set my username so I always have to awkwardly leave a chat message saying my name

This PR fixes that!

Disclaimer: this is technically untested, because staging API is hanging, but I edited the HTML locally on prod to produce the same effect, and it worked great.

### DIFF
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -220,7 +220,8 @@ export default class Chat extends Component {
         <EditableSpan
           ref={this.usernameInput}
           className="chat--username--input"
-          mobile={this.props.mobile}
+          // even when on mobile, we always want this to be editable
+          mobile={false}
           value={this.state.username}
           onChange={this.handleUpdateDisplayName}
           onBlur={this.handleBlur}
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -220,8 +220,6 @@ export default class Chat extends Component {
         <EditableSpan
           ref={this.usernameInput}
           className="chat--username--input"
-          // even when on mobile, we always want this to be editable
-          mobile={false}
           value={this.state.username}
           onChange={this.handleUpdateDisplayName}
           onBlur={this.handleBlur}
--- a/src/components/common/EditableSpan.js
+++ b/src/components/common/EditableSpan.js
@@ -86,24 +86,6 @@ export default class EditableSpan extends PureComponent {
     return new Caret(this.span.current && this.span.current.childNodes[0]);
   }
 
-  handleKeyDownMobile = (key) => {
-    const {caret} = this.state;
-    let newCaret = caret;
-    if (key === '{enter}') {
-      this.props.onPressEnter && this.props.onPressEnter();
-      return;
-    }
-    if (key === '{del}') {
-      this.text = this.text.substring(0, caret - 1) + this.text.substring(caret);
-      newCaret = caret - 1;
-    } else {
-      this.text = this.text.substring(0, caret) + key + this.text.substring(caret);
-      newCaret = caret + 1;
-    }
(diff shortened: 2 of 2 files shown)


tarehart:
Any idea why the contenteditable logic works this way? 
https://github.com/downforacross/downforacross.com/blob/20a420fa1ae1a39876145261aa53044f0ae744c1/src/components/common/EditableSpan.js#L142

The PR makes me nervous without that understanding.

Full disclosure, I'm not a maintainer so feel free to ignore, just trying to anticipate concerns Steven might have

stevenhao:
oh yeah i don't really remember how the content editable stuff works. i'll test this PR out in the vercel preview deployment, which seems to be working ok (it points to prod)

stevenhao:
cool so i verified that this fix works! but in terms of code style, i think it'd be better for contentEditable to just be true instead of passing mobile=false, since mobile=false when it's actually mobile is a bit confusing.

stevenhao:
patch file courtesy of https://chatgpt.com/share/67e1b6d5-2b68-8011-b6ef-8d02c307b6e7 

```
diff --git a/src/components/Chat/Chat.js b/src/components/Chat/Chat.js
index bf0ad7a..a858e64 100644
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -3,8 +3,6 @@
       {'You are '}
       <ColorPicker color={this.props.myColor} onUpdateColor={this.handleUpdateColor} />
       <EditableSpan ref={this.usernameInput} className="chat--username--input"
-        // even when on mobile, we always want this to be editable
-        mobile={false}
         value={this.state.username}
         onChange={this.handleUpdateDisplayName}
         onBlur={this.handleBlur}
diff --git a/src/components/common/EditableSpan.js b/src/components/common/EditableSpan.js
index a445473..d4d00b9 100644
--- a/src/components/common/EditableSpan.js
+++ b/src/components/common/EditableSpan.js
@@ -87,23 +87,6 @@ export default class EditableSpan extends PureComponent {
     return new Caret(this.span.current && this.span.current.childNodes[0]);
   }
 
-  handleKeyDownMobile = (key) => {
-    const {caret} = this.state;
-    let newCaret = caret;
-    if (key === '{enter}') {
-      this.props.onPressEnter && this.props.onPressEnter();
-      return;
-    }
-    if (key === '{del}') {
-      this.text = this.text.substring(0, caret - 1) + this.text.substring(caret);
-      newCaret = caret - 1;
-    } else {
-      this.text = this.text.substring(0, caret) + key + this.text.substring(caret);
-      newCaret = caret + 1;
-    }
-    this.props.onChange(this.text);
-    this.setState({caret: newCaret});
-  };
 
   handleKeyDown = (e) => {
     if (e.key === 'Tab') {
@@ -139,7 +122,7 @@ export default class EditableSpan extends PureComponent {
           style={style}
           className={`editable-span ${this.props.className || ''}`}
           ref={this.span}
-          contentEditable={this.props.mobile ? undefined : true}
+          contentEditable={true}
           onFocus={this.handleFocus}
           onBlur={this.handleBlur}
           onKeyDown={this.handleKeyDown}

```

cameronblandford:
yeah since i couldnt test, i wanted to introduce as little opportunity for side effects as possible. agreed that the patch seems better, given that everything works!

would also love to get connected to staging :)

cameronblandford:
@stevenhao done :+1:

stevenhao:
nice! re: staging, i think the server is down for some reason. but i think you should be able to connect directly to the prod backend server if you need to test a frontend change? not 100% sure if you need credentials or osmething.

stevenhao:
anyways, thanks for the contribution! i noticed the mobile username issue the other day too, so glad it's fixed now.

PR #345, opened 2025-03-25
Fix scroll bug

cameronblandford:
Context:
- Two users (A and B) are both looking at the clue list in a game
- User A scrolls all the way down to the bottom of the clue list without clicking anything
- User B clicks anywhere on the clue list

Expected behavior:
- Nothing should happen for User A
- User B should have the clue move 20% away from the top of the screen

Observed behavior:
- User A has their scroll position reset to their last-selected clue

---

I tried writing this a bunch of different ways as a class component, none of them worked. So I ended up opting for a standard functional component. As a result, the diff is bigger than I'd like. I haven't tested it with vim mode, but everything else functions as expected.

---

Note that the refactor into a functional component was 90% Claude's doing, I tested a bunch after but am operating partially blind.

### DIFF
--- a/server/model/pool.ts
+++ b/server/model/pool.ts
@@ -6,7 +6,9 @@ export const pool = new pg.Pool({
   user: process.env.PGUSER || process.env.USER,
   password: process.env.PGPASSWORD,
   database: process.env.PGDATABASE,
-  ssl: {
-    rejectUnauthorized: false,
-  }
+  ...(process.env.NODE_ENV === 'production' && {
+    ssl: {
+      rejectUnauthorized: false,
+    },
+  }),
 });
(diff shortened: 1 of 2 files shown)


cameronblandford:
cc @stevenhao

cameronblandford:
cc @stevenhao again, what do you think?

stevenhao:
taking a look!

server/model/pool.ts
```diff
@@ -6,7 +6,9 @@ export const pool = new pg.Pool({
   user: process.env.PGUSER || process.env.USER,
   password: process.env.PGPASSWORD,
   database: process.env.PGDATABASE,
-  ssl: {
-    rejectUnauthorized: false,
-  }
+  ...(process.env.NODE_ENV === 'production' && {
```
stevenhao, line 9: this looks safe, but what was the reason it was necessary again?

PR #346, opened 2025-05-20
win indicator for fencing

hermanmalik:
Hi! This is my first PR on github, let me know if I'm doing it right or if there's a test suite or something like that.

This commit just adds a simple win indicator for fencing because right now it's very annoying that the fencing will end and neither team knows who won or gets to see the board.

If you would prefer a bigger commit, I'm hoping to add some other improvements to fencing on this branch after I get free in a couple weeks or so, but I'd like to merge this in so I can have more fun playing fencing with my friends in the meantime.

The specific content of this fix is in `src/components/Fencing/FencingScoreboard.tsx`, where I added a winIndicator `span` next to the winning team name that says "🏆 Winner!". It displays only if every cell is filled.

### DIFF
--- a/src/components/Fencing/FencingScoreboard.tsx
+++ b/src/components/Fencing/FencingScoreboard.tsx
@@ -53,10 +53,17 @@ export const FencingScoreboard: React.FC<{
   );
 
   // Determine if the game is complete and which team won
+  // should be able to handle ties with any number of teams
   const isGameComplete = props.gameState.game?.grid.every((row) =>
     row.every((cell) => cell.good || cell.black)
   );
-  const winningTeam = isGameComplete ? _.maxBy(_.values(props.gameState.teams), 'score') : null;
+  const winningTeams = isGameComplete
+    ? (() => {
+        const teams = _.values(props.gameState.teams).filter(Boolean);
+        const maxScore = _.maxBy(teams, 'score')?.score;
+        return teams.filter((team) => team?.score === maxScore);
+      })()
+    : null;
 
   const teamData = _.keys(props.gameState.teams).map((teamId) => ({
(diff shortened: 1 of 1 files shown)


src/components/Fencing/FencingScoreboard.tsx
```diff
@@ -46,6 +51,13 @@ export const FencingScoreboard: React.FC<{
       Leave Team
     </button>
   );
+
+  // Determine if the game is complete and which team won
+  const isGameComplete = props.gameState.game?.grid.every((row) =>
+    row.every((cell) => cell.good || cell.black)
+  );
+  const winningTeam = isGameComplete ? _.maxBy(_.values(props.gameState.teams), 'score') : null;
```
brownie-in-motion, line 59: Maybe if more than one team is tied for first, they can all be winners? As is, it seems like ties would go to team 1.

hermanmalik:
Thank you @brownie-in-motion! I've fixed it by making a list of winning teams. This way if in the future fencing allows more than two teams it should be okay with arbitrary ties.

```
  const winningTeams = isGameComplete
    ? (() => {
        const teams = _.values(props.gameState.teams).filter(Boolean);
        const maxScore = _.maxBy(teams, 'score')?.score;
        return teams.filter((team) => team?.score === maxScore);
      })()
    : null;
```

stevenhao:
ok i tested on the preview link & managed to get to game end. seems like the winner thing shows up. in a similar vein to showing winner, it'd be nice to show overall board % completion & number of remaining unfilled squares.

image


unrelated to this PR probably, but fencing felt kinda buggy -- a lot of my typed cells didn't auto-check for some reason & stayed in pencil format. might've been a network issue, not sure.

PR #347, opened 2025-05-23
Don't flip clue when scrolling clue bar

a13o:
On Safari iOS, scrolling long clues will also flip the clue direction when you release the scroll. This makes it difficult to read the bottom of the clue and fill in an answer.

https://github.com/user-attachments/assets/d6d715c2-1980-44ea-b459-d29bdfc34e95

This PR contains a fix. Scrolling the clue will no longer flip the clue direction. Tapping the clue bar will still flip the clue direction. There's a buffer of 6 pixels for ergonomics.

https://github.com/user-attachments/assets/27dbd8f6-2e6a-49d8-ad09-096f24833d47

There was a lot of unused code in my way that I removed. This code appeared to be for an unfinished clue bar gestures feature. It was accessible via a `enableClueBarGestures` flag that was never set. I removed all code behind that flag as well as the flag, and kept the touchStart and touchEnd events for my bug fix.

### DIFF
--- a/src/components/Player/MobileGridControls.js
+++ b/src/components/Player/MobileGridControls.js
@@ -3,21 +3,15 @@ import './css/mobileGridControls.css';
 import React from 'react';
 import Flex from 'react-flexview';
 import {MdKeyboardArrowLeft, MdKeyboardArrowRight} from 'react-icons/md';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import classnames from 'classnames';
 import _ from 'lodash';
 import Clue from './ClueText';
 import GridControls from './GridControls';
 import GridObject from '../../lib/wrappers/GridWrapper';
-import * as gameUtils from '../../lib/gameUtils';
-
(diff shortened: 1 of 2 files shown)


stevenhao:
nice, thanks for fixing this & cleaning up the dead code. would be nice if we could make large clues visible without needing to scroll, if you have ideas for that i'd be down to review another PR!

a13o:
@stevenhao Thanks for the quick review. I did try some stuff to make the clue bar taller, but screen real estate is just at such a premium when the keyboard is open. The two lines of text height that’s there now is good for like 90% of clues. Although the puzzle’s big ‘gimmick’ clue is usually in the 10% so that’s frustrating…

Safari iOS doesn’t give a way to hide the autocomplete bar even if auto completions are disabled. And they add another bar with the domain name I guess for security reasons. Really annoying and wish we could hide those for more space!

I’ll keep thinking of ways to improve the mobile experience. That’s mainly how I use the site.

PR #350, opened 2025-06-09
reveal all cells when fencing game is done, and add confetti 

hermanmalik:
When you finish a fencing game it would be nice to be able to see the whole board, since right now the other team's portion stays grayed out.

I added an event called `revealAllClues` in `src/shared/fencingGameEvents` which makes all clues visible, and this automatically makes all cells visible. I originally tried using the existing `reveal` event and modifying the scope but this seemed simpler.

For the reveal logic in `Fencing.tsx`, I used `useEffect` instead of adding the if block to the existing code in `useUpdateEffect` because apparently it won't work on page load if it uses `useUpdateEffect`, but I'm unsure if this is correct.

I also moved the check for `isGameComplete` from `FencingScoreboard.tsx` to `Fencing.tsx`, so that I could use it for confetti and for the clue reveal.

### DIFF
--- a/src/components/Player/GridControls.js
+++ b/src/components/Player/GridControls.js
@@ -388,6 +388,10 @@ export default class GridControls extends Component {
   }
 
   typeLetter(letter, isRebus, {nextClueIfFilled} = {}) {
+    const {r, c} = this.props.selected;
+    if (!this.isSelectable(r, c)) {
+      return; // don't type in hidden/non-selectable cells
+    }
     if (this.props.beta) {
       return this.typeLetterSync(letter, isRebus, {nextClueIfFilled});
     }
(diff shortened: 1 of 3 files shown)


hermanmalik:
Added another commit to fix tab and arrow key navigation. I'm not sure how to make this a separate merge request on the same branch - should I have started a new branch for this commit?

It has been annoying (to me, at least) that navigation makes you select hidden cells, and you can even place letters in hidden cells, so I modified `selectNextClue` and `isSelectable` in GridControls.js to fix this. This should make it so you can only play with your area, and it makes fencing feel a fair bit more polished.

For `isClueSelectable` I just added a simple check `!this.props.grid[r][c].isHidden` for hidden squares. For `selectNextClue` it was just accepting `getNextClue` and selecting it, so I added a small method `isClueSelectable` and used the `safe_while` that was already there to keep calling `getNextClue` until we get a selectable one.

stevenhao:
sorry for the delay. skimmed the diff & tested lightly, will merge.

stevenhao:
thanks for contributing! if you have other improvements for fencing feel free to send them my way. tagging me in github is a good idea if i miss a notif.

PR #352, opened 2025-07-14
Improve mobile screen snapping

a13o:
This PR fixes a couple of issues demonstrated below

- There's a padding around the puzzle area which constricts the viewport
- The viewport snapping effect doesn't work on the y-axis

https://github.com/user-attachments/assets/dcd21d61-a415-43f4-8f59-9a2d9adcda42

Here is the new behavior after this PR

https://github.com/user-attachments/assets/2b8437a6-d831-476b-86b1-597ee71c7d5f

### DIFF
--- a/src/components/Game/Game.js
+++ b/src/components/Game/Game.js
@@ -22,7 +22,6 @@ export default class Game extends Component {
       listMode: false,
       pencilMode: false,
       autocheckMode: false,
-      screenWidth: 0,
       vimMode: false,
       vimInsert: false,
       vimCommand: false,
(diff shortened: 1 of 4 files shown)


src/components/Game/Game.js
```diff
@@ -267,7 +264,7 @@ export default class Game extends Component {
       clues[dirToHide] = _.assign([], clues[dirToHide]).map((val) => val && '-');
     }
     const opponentGrid = this.opponentGame && this.opponentGame.grid;
-    const {screenWidth} = this.state;
+    const screenWidth = window.innerWidth - 1; // this is important for mobile to fit on screen
```
a13o, line 267: This change isn't strictly necessary; but it avoids an initial render of the grid on mobile where all the grid cells are 1 px large. That would happen because screenWidth wasn't computed until `componentDidMount`, so the initial calculation ran as if `screenWidth = 0`.

Since nothing else ever modifies `screenWidth`, and it's a cheap calculation, there's no need to cache it.

a13o:
@stevenhao can you review? thank you

stevenhao:
Looks good! Thanks for fixing

